PR #37, opened 2015-01-08
Remove dependency of Sufia-Models and upgrade to Fedora-4

jcoyne:


scherztc:
I follow removing the dependency on sufia-models and refactoring the generic_work_rdf_properties data stream to be inline with Fedora 4.  With curate we had a few other properties that may not be part of the base GenericWork but I wanted to point them out.  property :extent, property :rights, and property: source.  Are those needed here?

scherztc:
I also understand the need to account for DC Metadata as the baseline in the GenericWork class and allow for extending.  I see that we have: property :part_of, predicate: RDF::DC.isPartOf but how do we deal with : map.part(:to => "hasPart", in: RDF::DC)?

mjgiarlo:
:+1:

jcoyne:
@scherztc I don't have any reservations with you extending the vocabulary, I'm just using what was already built, I just moved it from one path to another.

scherztc:
Thanks.  I will work on a PR for adding those three fields to the vocabulary.  In regards to my other question about   : map.part(:to => "hasPart", in: RDF::DC)?  Is that getting discussed over here?  https://github.com/projecthydra-labs/hydra-works/issues/38

PR #45, opened 2015-02-12
Changing model name from GenericFile to File to match documentation

cam156:

PR #46, opened 2015-02-12
Splitting with generic files and the cleanup so we can just with in scen...

cam156:
...arios that we do not want to cleanup the files too


cam156:
@jcoyne Are you talking about the previous PR?

cam156:
this one? https://github.com/projecthydra-labs/hydra-works/pull/45

scherztc:
Sorry.  I read it as, we were doing something similar with files that we did with works... A base file model and an extended GenericFile.

PR #48, opened 2015-02-12
Splitting with generic files and the cleanup so we can just with in scen...

cam156:
...arios that we do not want to cleanup the files too


jcoyne:
Shouldn't we check if the GenericFiles are pointed at by other GenericWorks?

jcoyne:
This doesn't make sense to do anymore.

Issue #73, opened 2015-05-11
Implement Thumbnail file type


awead:
Relates to projecthydra/sufia#1010

Build a thumbnail file class in HydraWorks such that an implementation could do:
``` ruby
class MyGenericFile < Hydra::Works::GenericFile
  has_one :thumbnail, predicate: ::RDF::URI(http://hydraworks.namespace/hasThumbnail), class_name: "HydraWorks::Thumbnail"
end
```
and triples associated with MyGenericFile would include `<> hasThumbnail HydraWorks::Thumbnail`

The thumbnail class extends Hydra::PCDM::File and has its own rdf type as per https://github.com/duraspace/pcdm/pull/5

escowles:
See https://github.com/duraspace/pcdm/pull/7 -- adding a file use vocabulary as a PCDM extension, with RDF types and predicates for the identified file use terms.

jcoyne:
I'm not sure why we'd need a class for this. Wouldn't a URI suffice?

awead:
@jcoyne I could see that. I was just putting up an example to facilitate discussion.

escowles:
This should probably be an AF::File with rdf:type=pcdmuse:ThumbnailImage added to the RDF description like in https://github.com/awead/sufia-pcdm/blob/master/pcdm-thumbnail.ru

awead:
@escowles heh. just did that. see #79. I haven't added a property to the Work that points directly to the file, instead, it's just querying all the hasFile members and returning those with the pcdm.Thumbnail type.

awead:
fixed in #105

PR #79, opened 2015-05-12
Implement adding original files, thumbnails and extracted text to Hydra::Works

awead:
This is an initial implementation that addresses #73. It adds the `files` direct container to Works and provides a mechanism for designating one of them as a pcdm.Thumbnail.

### DIFF
--- a/lib/hydra/works/models/concerns/contained_files.rb
+++ b/lib/hydra/works/models/concerns/contained_files.rb
@@ -1,0 +1,28 @@
+module Hydra::Works::ContainedFiles
+  extend ActiveSupport::Concern
+
+  # We can use PCDM vocab class when projecthydra-labs/hydra-pcdm#80 is merged
+
+  included do
+    directly_contains :files, has_member_relation: ::RDF::URI("http://pcdm.org/hasFile"), class_name: "Hydra::Works::File"
+  end
+
+  def original_file
+    self.files.reject do |file|
+      file.metadata_node.query(predicate: RDF.type, object: ::RDF::URI("http://pcdm.org/OriginalFile")).map(&:object).empty?
+    end
+  end
+
+  def thumbnail
+    self.files.reject do |file|
+      file.metadata_node.query(predicate: RDF.type, object: ::RDF::URI("http://pcdm.org/ThumbnailImage")).map(&:object).empty?
+    end
+  end
+
+  def extracted_text
+    self.files.reject do |file|
+      file.metadata_node.query(predicate: RDF.type, object: ::RDF::URI("http://pcdm.org/ExtractedText")).map(&:object).empty?
+    end
+  end  
+
+end
--- a/lib/hydra/works/models/file.rb
+++ b/lib/hydra/works/models/file.rb
@@ -1,0 +1,31 @@
+module Hydra::Works
+  class File < Hydra::PCDM::File
+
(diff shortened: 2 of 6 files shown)


lib/hydra/works/models/concerns/contained_files.rb
```diff
@@ -0,0 +1,17 @@
+module Hydra::Works::ContainedFiles
+  extend ActiveSupport::Concern
+
+  included do
+    directly_contains :files, has_member_relation: ::RDF::URI("http://pcdm.org/hasFile"), class_name: "Hydra::Works::File"
```
jcoyne: Can you use the PCDM vocabulary class?

lib/hydra/works/models/concerns/contained_files.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def thumbnail
+    files = []
```
jcoyne: files is a confusing name for a local variable since we already have a files method that we're calling on line 10? how about `matches`?

lib/hydra/works/models/concerns/contained_files.rb
```diff
@@ -0,0 +1,17 @@
...
+    self.files.each do |file|
```
jcoyne: Using `self.files.to_a.select` will be more concise.

lib/hydra/works/models/concerns/contained_files.rb
```diff
@@ -0,0 +1,17 @@
...
+      files << file unless file.metadata_node.query(predicate: RDF.type, object: ::RDF::URI("http://pcdm.org/#Thumbnail")).map(&:object).empty?
```
jcoyne: Move this to Hydra::PCDM::File.thumbnail?

lib/hydra/works/models/concerns/contained_files.rb
mjgiarlo: Can you refactor this method using `#reject_if` or similar?

lib/hydra/works/models/file.rb
```diff
@@ -0,0 +1,35 @@
+module Hydra::Works
+  class File < Hydra::PCDM::File
+
+    # Need this here until projecthydra-labs/hydra-pcdm#77 is closed
+    metadata do
+      configure type: RDFVocabularies::PCDMTerms.File
```
jcoyne: Can we update ActiveTriples to support multiple types. That would greatly simplify this code.

lib/hydra/works/models/concerns/contained_files.rb
awead: @jcoyne what about Hydra::Works::File.thumbnail? I didn't think thumbnails would be a part of hydra-pcdm... only hydra-works.

mjgiarlo: I don't think PCDM knows anything other than Collection, Object, and File (and proxies).  Right?  Thumbnails are a higher-layer concern.

awead: @mjgiarlo see projecthydra-labs/hydra-pcdm#80. HP already has the RDFVocabularies class in it to which these terms would be added.

jcoyne:
I think we should re-think this. I don't think we want a special file class (Works::File).  There's no reason to not use service objects to add the thumnail, etc.

PR #82, opened 2015-05-12
Moves a generic file from one generic work to another. 

### LINKED ISSUE #76: service object: MoveGenericFileToGenericWork
```
module Hydra::Works
  class MoveGenericFileToGenericWork

    ##
    # Move a generic file from one generic work to another
    #
    # @param [Hydra::Works::GenericWork] :old_parent_work where the generic file currently lives
    # @param [Hydra::Works::GenericWork] :new_parent_work where the generic file is being moved
    # @param [Hydra::Works::Generic_File] :child_generic_file being moved
    #
    # @return [Hydra::Works::GenericWork] the destination generic work

    def self.call( old_parent_work, new_parent_work, child_generic_file )
        # TODO validate all parameters
        # TODO create association between new_parent_work and child_generic_file
        # TODO remove association between old_parent_work and child_generic_file

        # NOTE: This is not a copy.  No new fedora objects should be created as part of this process.
  end
end
```

hectorcorrea:
Fixes #76

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ spec/internal
 *.a
 mkmf.log
 jetty
+.DS_Store
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -41,6 +41,7 @@ module Hydra
     autoload :GetGenericWorksFromGenericWork,      'hydra/works/services/generic_work/get_generic_works'
     autoload :GetGenericFilesFromGenericWork,      'hydra/works/services/generic_work/get_generic_files'
     autoload :GetRelatedObjectsFromGenericWork,    'hydra/works/services/generic_work/get_related_objects'
+    autoload :MoveGenericFileToGenericWork,        'hydra/works/services/generic_work/move_generic_file'
     autoload :RemoveGenericWorkFromGenericWork,    'hydra/works/services/generic_work/remove_generic_work'
     autoload :RemoveGenericFileFromGenericWork,    'hydra/works/services/generic_work/remove_generic_file'
     autoload :RemoveRelatedObjectFromGenericWork,  'hydra/works/services/generic_work/remove_related_object'
@@ -63,7 +64,6 @@ module Hydra
     # file services
     autoload :AddTypeToFile,                       'hydra/works/services/file/add_type'
     autoload :GetMimeTypeForFile,                  'hydra/works/services/file/get_mime_type'
-
 
     # model validations
     def self.collection? collection
--- a/lib/hydra/works/services/generic_work/move_generic_file.rb
+++ b/lib/hydra/works/services/generic_work/move_generic_file.rb
@@ -1,0 +1,19 @@
+module Hydra::Works
+  class MoveGenericFileToGenericWork
+
+    ##
+    # Move a generic file from one generic work to another
+    #
+    # @param [Hydra::Works::GenericWork] :old_parent_work where the generic file currently lives
+    # @param [Hydra::Works::GenericWork] :new_parent_work where the generic file is being moved
+    # @param [Hydra::Works::GenericFile] :child_generic_file being moved
+    #
+    # @return [Hydra::Works::GenericWork] the destination generic work
+    def self.call(old_parent_work, new_parent_work, child_generic_file)
+      AddGenericFileToGenericWork.call(new_parent_work, child_generic_file)
+      RemoveGenericFileFromGenericWork.call(old_parent_work, child_generic_file)
+      new_parent_work
+    end
+
+  end
+end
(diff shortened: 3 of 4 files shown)


hectorcorrea:
@elrayle could you take a look at this PR and see if this what you had in mind?

lib/hydra/works/services/generic_work/move_to_generic_file.rb
```diff
@@ -0,0 +1,34 @@
...
+    #
+    # @param [Hydra::Works::GenericWork] :old_parent_work where the generic file currently lives
+    # @param [Hydra::Works::GenericWork] :new_parent_work where the generic file is being moved
+    # @param [Hydra::Works::GenericFile] :child_generic_file being moved
+    #
+    # @return [Hydra::Works::GenericWork] the destination generic work
+    def self.call(old_parent_work, new_parent_work, child_generic_file)
+      raise ArgumentError, "old_parent_work must be a Hydra::GenericWork" unless old_parent_work.is_a?(Hydra::Works::GenericWork)
+      raise ArgumentError, "new_parent_work must be a Hydra::GenericWork" unless new_parent_work.is_a?(Hydra::Works::GenericWork)
+      raise ArgumentError, "child_generic_file must be a Hydra::GenericFile" unless child_generic_file.is_a?(Hydra::Works::GenericFile)
```
awead: change these to use the newly added Hydra::Works.generic_work? and .generic_file? methods.

hectorcorrea: Done.

lib/hydra/works/services/generic_work/move_to_generic_file.rb
```diff
@@ -0,0 +1,37 @@
...
+      remove_file_from_work old_parent_work, child_generic_file
+      new_parent_work.reload
+    end
+
+    private
+
+    def self.add_file_to_work(work, file) 
+      # TODO: This should be changed to work.generic_files once 
+      #       generic_files has been fully implemented in Hydra::Works
+      work.members << file
```
no-reply: This, and `work.members.delete file` below, are Law of Demeter violations. What do people think about `work.add_member(file)` and `work.delete_member(file)` ?

hectorcorrea: @no-reply I like your idea which coincidentally matches the approach that @elrayle was suggesting during the sprint call earlier today.

no-reply: @elrayle: :fireworks:

Awesome. Keeping close watch for things like this should help ServiceObjects stay genuinely decoupled. :+1:

elrayle:
With CRUD services, this becomes...
```
module Hydra::Works
  class MoveGenericFileToGenericWork
    def self.call(old_parent_work, new_parent_work, child_generic_file)
      AddGenericFileToGenericWork( new_parent_work, child_generic_file )
      RemoveGenericFileFromGenericWork( old_parent_work, child_generic_file  )
    end 
  end
end
```

AddGenericFileToGenericWork is part of PR https://github.com/projecthydra-labs/hydra-works/pull/106
RemoveGenericFileFromGenericWork is coming in the next day or two.  I'd like to get PR 106 merged first to avoid a bunch of rebasing.

elrayle:
@hectorcorrea The services for Add and Remove have been merged to master.  You should be able to use those for the Move service.  See previous comment with sample code.  You don't need to validate your incoming parameters because the Add and Remove services will do that for you.

hectorcorrea:
@elrayle Done.

hectorcorrea:
@jeremyf Mocks won't work in this case because the underlying code validates to make sure the arguments are of a certain type (i.e. downstream the code executes something along the lines of  Hydra::Works.generic_work? workA)

jcoyne:
@hectorcorrea Those functions or the type they expect could be mocked out too.

jcoyne:
For example: 
```ruby
workA = double(type: [RDFVocabularies::PCDMTerms.Work])
```

hectorcorrea:
@jcoyne @jeremyf  Two things. (1) I refactored the tests to use doubles as you suggested. I had to introduce a dummy variable to force RSpec to recognize "RDFVocabularies::PCDMTerms". Any idea on to prevent this dummy variable? 

(2) Although this test is now faster because the use of doubles, I find the new test less than ideal, and to be honest downright horrifying. The test now knows and depends on (a) the internal calls that the code under test makes, (b) the order of the arguments of these internal calls, and (c) the way types are validated in these internal calls. I am having a hard time calling this "a better" test than the original, despite the fact that is a faster test.

elrayle:
@hectorcorrea I agree with you that the test is less clear using the fake objects.  And, if the definition of the validation process changes, these tests will fail.  I don't see how the actual service is being tested.  It is called in the last line of the test, but there is no verification that a) file1 is a member of workB, b) file1 is not a member of workA.

Given the general lack of documentation in Hydra, tests also serve the purpose of demonstrating how to use a function.  This test does not do that.  I know how it works and find the test hard to follow.

hectorcorrea:
@elrayle thanks for the feedback. I put back the original test.

Issue #75, opened 2015-05-11
Use ActiveTriples mechanism for setting multiple types when available


elrayle:
Effects:
* collection_behavior.rb
* work_behavior.rb
* file_behavior.rb

Replace the following code:   (substitute appropriate type for WorksTerms.Collection)
```
    def initialize(*args)
      super(*args)

      t = get_values(:type)
      t << RDFVocabularies::WorksTerms.Collection
      set_value(:type,t)
    end
```
With something like...   (TBD depending on ActiveTriples implementation)
```
    included do
      type RDFVocabularies::WorksTerms.Collection
    end
```

elrayle:
Fixed by PR https://github.com/projecthydra-labs/hydra-works/pull/101

PR #101, opened 2015-05-13
add multiple types for collections, generic works, and generic files

elrayle:
NOTE: The Gemfile ties active_fedora and active-triples to specific branches that allows for multiple types.


elrayle:
Fixes https://github.com/projecthydra-labs/hydra-works/issues/75

PR #105, opened 2015-05-15
Identify file types and upload an original file

awead:
This supports identifying original file, thumbnail and extracted text files that are attached to a GenericFile. It also adds services for uploading a file to a generic file and having its thumbnail generated via hydra-derivatives.


elrayle:
Looks good, except for one extraneous file...
               lib/hydra/works/services/collection/.keep

Issue #104, opened 2015-05-14
Services:  Add/Get services for Collections, GenericWorks, GenericFiles


elrayle:
Follow same pattern as in pcdm services.

elrayle:
Reference for pcdm services:  https://github.com/projecthydra-labs/hydra-pcdm/pull/93

elrayle:
Fixed by PR https://github.com/projecthydra-labs/hydra-works/pull/106

PR #106, opened 2015-05-15
New services for adding and getting collections, generic works, and generic files.  De-dup tests.

elrayle:


awead:
@elrayle It would be helpful if the tests marked with `xit` used skip instead:
```
it "should do somthing" do
  skip "skipping this test because issue #XX needs to be addressed"
end
```
The skip messages are displayed when the test suite is run so we can why we're not running it. Please make sure that any skipped tests have a corresponding issue. I think this is the case already, but we need to indicate which test go with which issues.

Other than, :clap: great work!!!!

elrayle:
@awead I updated the PR to use skip instead of xit.  That was a good suggestion.  I like having the issue number print out when the tests run.  Similarly, when I have a TODO in the code, I include the issue number in the TODO statement.

elrayle:
Once this is merged, I will follow it with a PR related to remove services.

elrayle:
Fixes https://github.com/projecthydra-labs/hydra-works/pull/104

PR #114, opened 2015-05-19
Remove hydra-head and hydra-collections as dependencies. Fixes #99

### LINKED ISSUE #99: Remove dependencies that require Rails (hydra-head, hydra-collections)
If possible, remove dependencies that require Rails (hydra-head, hydra-collections)

Our plan was that hydra-works should implement hydra-pcdm and should not depend on Rails. The hydra-works.gemspec does not currently match with that intention.  It has hydra-head (which requires rails) and hydra-collections as dependencies.  https://github.com/projecthydra-labs/hydra-works/blame/master/hydra-works.gemspec#L21-L22 

If removing hydra-head and hydra-collections breaks the gem, record tickets indicating what has to change in those gems in order to allow hydra-works to function without Rails dependencies. 

**Example 1**: the ACL part of hydra-access-controls needs to become its own gem that doesn't rely on Rails
**Example 2**: and hydra-collections models need to live in a hydra-collections-models

jcoyne:


mjgiarlo:
:+1: if travis is :relaxed:

PR #117, opened 2015-05-28
refactors namespaces to match Code Shredding document

flyingzumwalt:
See [Code Shredding document](https://docs.google.com/spreadsheets/d/1uyCtjcBeSr0rWu4AieVplygrhWVTwwR0v6Itq6PBeA0/edit)

* Creates a Hydra::Works::File namespace and puts File-specific modules in there 
* Moves `generic_files` accessors into an AggregatesGenericFiles module
* adds some documentation

Also refactors GenericWork and GenericFile to be namespaces rather than classes; Renames `GenericWork` and `GenericFile` classes to `GenericWork::Base` and `GenericFile::Base`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gemspec
 
 gem 'activefedora-aggregation', github: 'projecthydra-labs/activefedora-aggregation'
 gem 'active-fedora', github: 'projecthydra/active_fedora', branch: 'master'
-gem 'hydra-pcdm', github: 'projecthydra-labs/hydra-pcdm'
+gem 'hydra-pcdm', github: 'projecthydra-labs/hydra-pcdm', ref:'c8a4654'
 gem 'slop', '~> 3.6' # For byebug
 
 unless ENV['CI']
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -19,9 +19,10 @@ module Hydra
     autoload :CollectionBehavior,     'hydra/works/models/concerns/collection_behavior'
     autoload :GenericWorkBehavior,    'hydra/works/models/concerns/generic_work_behavior'
     autoload :GenericFileBehavior,    'hydra/works/models/concerns/generic_file_behavior'
-    autoload :ContainedFiles,         'hydra/works/models/concerns/contained_files'
-    autoload :Derivatives,            'hydra/works/models/concerns/derivatives'
-    autoload :MimeTypes,              'hydra/works/models/concerns/mime_types'
+    autoload :AggregatesGenericFiles, 'hydra/works/models/concerns/aggregates_generic_files'
+    autoload :AggregatesGenericWorks, 'hydra/works/models/concerns/aggregates_generic_works'
+    autoload :AggregatesCollections,  'hydra/works/models/concerns/aggregates_collections'
+    autoload :GenericFile,            'hydra/works/models/generic_file'
 
     # collection services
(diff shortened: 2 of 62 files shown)


lib/hydra/works/models/concerns/file.rb
```diff
@@ -0,0 +1,12 @@
+module Hydra::Works
+  # Namespace for Modules with functionality specific to Files
+  module File
```
flyingzumwalt: The equivalent namespace in sufia was Sufia::GenericFile, but in hydra-works, Hydra::Works::GenericFile is a class so we can't also use it as a namespace. 

One option: name the class Hydra::Works::GenericFileBase, which would allow us to call this module Hydra::Works::GenericFile instead of Hydra::Works::File.

awead:
:+1:

lib/hydra/works/models/concerns/file.rb
flyingzumwalt: @elrayle do you have an opinion about this?

mjgiarlo: How about `Hydra::Works::GenericFile::Base`?

awead: I'm not sure I like this. It destroys the parallels between Hydra::PCDM::File, which is a class.

flyingzumwalt:
@elrayle @awead @jcoyne at @mjgiarlo's suggestion I made a version that refactors GenericWork and GenericFile to be namespaces rather than classes & (to make that possible) renames the `GenericWork` and `GenericFile` classes to `GenericWork::Base` and `GenericFile::Base`

If everyone agrees with this approach, I will use the same pattern in the curation_concerns gem.

awead:
I'm getting confused now. Is Files no longer a namespace? Everything is under the GenericFile namespace now, yes?

What's the advantage to namespacing GenericFile, versus keeping it a class that includes one behavior module that in turn includes all the additional modules?

awead:
OK, I think I'm on-board with this. Are you trying to borrow the ActiveFedora::Base layout? So I'm seeing:
``` ruby
module Hydra::Works::GenericFile
  class Base
    include Derivatives, Characterization, BlahBlah
  end
end
```
I'll point out, that for the implementor, this implies:
``` ruby
class MyGenericFile < Hydra::Works::GenericFile::Base
end
```
as opposed to
``` ruby
class MyGenericFile < ActiveFedora::Base
  include Hydra::Works::GenericFile::Base
end
```
I guess I make more sense of these changes when I can see examples of what rearranging modules and class actually gets us. But I do find `MyGenericFile < Hydra::Works::GenericFile::Base` attractive.

jcoyne:
I have no opinion.

elrayle:
I don't see any problems with the change.  I assume this makes the code more consistent, and that is the reason for changing.  I am all for consistency as it makes code easier to read and maintain.  With that in mind, if this is accepted, I would recommend that you do something similar for GenericWork to move generic_works= and generic_works methods to aggregates_generic_works.rb and then generic_work_behavior and collection_behavior can include it.  Similar changes could be made at the pcdm level for collections and objects and at the hydra works level for collections (which may be just to include the pcdm aggregates_collections.rb).

flyingzumwalt:
I made the Aggregates* stuff consistent (good suggestion @elrayle) and squashed the commits.  Ready to merge?

awead:
:+1: again, after refactor.

PR #122, opened 2015-06-02
Refactor AddOriginalFile call to take object, file, options.

grosscol:
*This will break existing calls to AddOriginalFile*
Add spec for AddOriginalFile service.
Require file arg to AddOriginalFile to respond to read.
Remove attribute option defaults from AddOriginalFile.
Refactor uses of AddOriginalFile.
Convert path in calls to AddOriginalFile to file using File.open(path)
Do attribute option discovery in UploadFileToGenericFile.


lib/hydra/works/services/generic_file/add_original_file.rb
```diff
@@ -1,25 +1,28 @@
 module Hydra::Works
   class AddOriginalFile
 
-    # Adds the original file to the work
+    # Adds the original file to the object (e.g. GenericFile)
+    # :original_name
```
jcoyne: Please use YARD documentation. See https://github.com/projecthydra/hydra/blob/master/CONTRIBUTING.md#documenting-code

lib/hydra/works/services/generic_file/upload_file.rb
```diff
@@ -6,11 +6,17 @@ def self.call(object, path, additional_services=[], replace=false)
...
     
+      opts = { 
+        original_name: ::File.basename(path), 
+        mime_type: Hydra::Works::GetMimeTypeForFile.call(path),
+        replace: replace
+      }
+
       if object.original_file.nil?
-        Hydra::Works::AddOriginalFile.call(object, path)
+        Hydra::Works::AddOriginalFile.call(object, File.open(path), opts)
```
jcoyne: This may not be germain to this PR, but why doesn't UploadFileToGenericFile take a File rather than a Path?

grosscol: I started to go through and change the other services, but limited my scope to the stated issue.  The thought was that could be a simple issue for a newer contributor to refactor in a similar fashion.

lib/hydra/works/services/generic_file/add_original_file.rb
grosscol: Will do.

lib/hydra/works/services/generic_file/upload_file.rb
jcoyne: After thinking about it for a few moments, I believe its' just the classes name that is proving confusing to me.  Something more like "UploadLocalFile" would be more clear to me.

lib/hydra/works/services/generic_file/add_original_file.rb
```diff
@@ -1,10 +1,12 @@
 module Hydra::Works
   class AddOriginalFile
 
-    # Adds the original file to the work
+    # Convenience service for attaching original file
+    # @see Hydra::Works::AddFileToGenericFile 
 
-    def self.call(object, path, replace=false)
-      Hydra::Works::AddFileToGenericFile.call(object, path, ::RDF::URI("http://pcdm.org/OriginalFile"), replace)
+    def self.call(object, file, options={})
```
jcoyne: Please document parameters with YARD

grosscol: Should the comment block from AddFileToGenericFile be duplicated here?  I was unsure if the `@see` tag would be sufficient. Should it read `# For parameter descriptions, @see Hydra::Works::AddFileToGenericFile` or just copy paste the entire comment block from AddFileToGenericFile?

grosscol:
Stale.  Going to start over with this since the code base has shifted significantly.

Issue #130, opened 2015-06-19
PersistDerivativesOutputFile Service


flyingzumwalt:
Implement Hydra::Works::PersistDerivativesOutputFile service

* `#call` conforms to the interface of hydra-derivatives PersistOutputFileService
* calls `Hydra::Works::AddFileToGenericFile` service

mjgiarlo:
Implemented in #132.

mjgiarlo:
This is mostly done, with some remaining work.  Closing this issue, and @flyingzumwalt will ticket the outstanding stuff.

PR #132, opened 2015-06-25
Persist dervivatives output service.

kevinreiss:
For #130. @flyingzumwalt please review. Do not merge until https://github.com/projecthydra/hydra-derivatives/pull/61 is sorted out.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gemspec
 gem 'activefedora-aggregation', github: 'projecthydra-labs/activefedora-aggregation', ref: '0dfe4f4'
 gem 'active-fedora', github: 'projecthydra/active_fedora', ref: 'caadc73'
 gem 'hydra-pcdm', github: 'projecthydra-labs/hydra-pcdm', ref: '11f46014'
+gem 'hydra-derivatives', github: 'projecthydra/hydra-derivatives', ref: '7a8377c'
 gem 'slop', '~> 3.6' # For byebug
 
 unless ENV['CI']
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -58,6 +58,7 @@ module Hydra
     autoload :RemoveRelatedObjectFromGenericFile,  'hydra/works/services/generic_file/remove_related_object'
     autoload :UploadFileToGenericFile,             'hydra/works/services/generic_file/upload_file'
     autoload :GenerateThumbnail,                   'hydra/works/services/generic_file/generate/thumbnail'
+    autoload :PersistDerivative,                   'hydra/works/services/generic_file/persist_derivative'
 
     # model validations
     def self.collection? collection
--- a/lib/hydra/works/services/generic_file/persist_derivative.rb
+++ b/lib/hydra/works/services/generic_file/persist_derivative.rb
@@ -1,0 +1,26 @@
+require 'hydra/derivatives'
+
+module Hydra::Works
+  class PersistDerivative < Hydra::Derivatives::PersistOutputFileService
+
+    ##
+    # Persists a derivative to a GenericFile
+    # This Service conforms to the signature of `Hydra::Derivatives::PersistOutputFileService`.  
+    # The purpose of this Service is for use as an alternative to the default Hydra::Derivatives::PersistOutputFileService.  It's necessary because the default behavior in Hydra::Derivatives assumes that you're using LDP Basic Containment. Hydra::Works::GenericFiles use IndirectContainment.  This Service handles that case.
+    # 
+    # @param [Hydra::Works::GenericFile::Base] object the file will be added to
+    # @param [File] file the dervivative filestream
+    # @param [String] destination_name path to the file
+    # @option opts [RDF::URI or String] type URI for the RDF.type that identifies the file's role within the generic_file
+    # @option opts [Boolean] update_existing when set to false, always adds a new file.  When set to true, performs a create_or_update
+    # @option opts [Boolean] versioning whether to create new version entries
+
+    def self.call(object, file, destination_name, opts={})
+      type = opts.fetch(:type, ::RDF::URI("http://pcdm.org/use#ServiceFile"))
+      update_existing = opts.fetch(:update_existing, true)
+      versioning = opts.fetch(:versioning, true)
+      Hydra::Works::AddFileToGenericFile.call(object, destination_name, type, update_existing: update_existing, versioning: versioning)
+    end
+
+  end
+end
--- a/spec/hydra/works/services/generic_file/persist_derivative_spec.rb
+++ b/spec/hydra/works/services/generic_file/persist_derivative_spec.rb
@@ -1,0 +1,23 @@
+require 'spec_helper'
+
+describe Hydra::Works::PersistDerivative do
+
+  let(:generic_file)        { Hydra::Works::GenericFile::Base.create }
+  let(:filename)            { "sample-file.pdf" }
+  let(:destination_name)    { File.join(fixture_path, filename) }
+  let(:file)                { File.new(destination_name) }
+  let(:type)                { ::RDF::URI("http://pcdm.org/use#ExtractedText") }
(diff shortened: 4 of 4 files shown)


lib/hydra/works/services/generic_file/persist_derivatives_output_file.rb
```diff
@@ -0,0 +1,22 @@
...
+
+module Hydra::Works
+  class PersistDerivativesOutputFile < Hydra::Derivatives::PersistOutputFileService
+
+    # Persists a Derivative to a GenericFile
+    # @param [Hydra::Works::GenericFile::Base] object the file will be added to
+    # @param [File] file the dervivative filestream
+    # @param [String] destination_name path to the file
+    # @option opts [RDF::URI or String] type URI for the RDF.type that identifies the file's role within the generic_file
+    # @option opts [Boolean] update_existing when set to false, always adds a new file.  When set to true, performs a create_or_update
```
kevinreiss: @flyingzumwalt do we need to allow these flags for AddFileToGenericFile to be overridden in this service method signature? Or should we just select appropriate defaults for derivative files?

grosscol: I would urge to permit the overrides.  I can envision use cases for not wanting versioning.  I'm not sure about use cases for update_existing being false.

flyingzumwalt:
@kevinreiss  I updated the af_is_broken branch to use the current hydra-works/master and latest active-fedora. We'll see what Travis says. The PR: #133

If the failures don't come up on that branch, then the bugs you're getting are actually bus in hydra-derivatives that should be fixed there before you merge this.

kevinreiss:
@flyingzumwalt this now now rebased againts hydra-works master. It is pegged to the latest master https://github.com/projecthydra/hydra-derivatives/commit/7a8377c12ba1697f825cef9f5b2cd3ec7b844228 commit on hydra-derivatives.

lib/hydra/works/services/generic_file/persist_derivatives_output_file.rb
```diff
@@ -0,0 +1,22 @@
+require 'hydra/derivatives'
+
+module Hydra::Works
+  class PersistDerivativesOutputFile < Hydra::Derivatives::PersistOutputFileService
+
+    # Persists a Derivative to a GenericFile
+    # @param [Hydra::Works::GenericFile::Base] object the file will be added to
```
flyingzumwalt: Could you add a comment explaining the purpose of the service?  Something like:
```
This Service conforms to the signature of `Hydra::Derivatives::PersistOutputFileService`.  
The purpose of this Service is for use an alternative to the default Hydra::Derivatives::PersistOutputFileService.  It's necessary because the default behavior in Hydra::Derivatives assumes that you're using LDP Basic Containment.  Hydra::Works::GenericFiles use IndirectContainment.  This Service handles that case.
```

lib/hydra/works/services/generic_file/persist_derivatives_output_file.rb
```diff
@@ -0,0 +1,22 @@
...
+    # @param [File] file the dervivative filestream
+    # @param [String] destination_name path to the file
+    # @option opts [RDF::URI or String] type URI for the RDF.type that identifies the file's role within the generic_file
+    # @option opts [Boolean] update_existing when set to false, always adds a new file.  When set to true, performs a create_or_update
+    # @option opts [Boolean] versioning whether to create new version entries (only applicable if +type+ corresponds to a versionable file)
```
jcoyne: does the 'only applicable...' still make sense?

flyingzumwalt: Good point. I'd re-word the description for that parameter.

lib/hydra/works/services/generic_file/persist_derivatives_output_file.rb
```diff
@@ -0,0 +1,26 @@
+require 'hydra/derivatives'
+
+module Hydra::Works
+  class PersistDerivativesOutputFile < Hydra::Derivatives::PersistOutputFileService
```
mjgiarlo: Though I wouldn't hold up the PR for this alone, I might like to see the class (and file) renamed more succinctly: `PersistDerivative`. That also reflects the description of the class on line 7 of this file.

lib/hydra/works/services/generic_file/persist_derivatives_output_file.rb
mjgiarlo: :+1: to allowing these to be tweaked via params.

lib/hydra/works/services/generic_file/persist_derivatives_output_file.rb
```diff
@@ -0,0 +1,26 @@
...
+    # 
+    # @param [Hydra::Works::GenericFile::Base] object the file will be added to
+    # @param [File] file the dervivative filestream
+    # @param [String] destination_name path to the file
+    # @option opts [RDF::URI or String] type URI for the RDF.type that identifies the file's role within the generic_file
+    # @option opts [Boolean] update_existing when set to false, always adds a new file.  When set to true, performs a create_or_update
+    # @option opts [Boolean] versioning whether to create new version entries
+
+    def self.call(object, file, destination_name, opts={})
+      type = opts[:type] || ::RDF::URI("http://pcdm.org/ServiceFile")
```
mjgiarlo: I believe the URI here should be `http://pcdm.org/use#ServiceFile` per https://github.com/duraspace/pcdm/blob/master/pcdm-ext/use.rdf#L50-L57

lib/hydra/works/services/generic_file/persist_derivatives_output_file.rb
```diff
@@ -0,0 +1,26 @@
...
+      update_existing = opts[:update_existing] || true
+      versioning = opts[:versioning] || true
```
mjgiarlo: For lines 19-21, I would recommend using the `#fetch` method defined on hashes rather than the `||` operator because if, let's say, I set `opts[:versioning]` to `false`, I get weird behavior:

```ruby
opts[:type] = false
# => false
opts[:type] || true
# => true
# whereas:
opts.fetch(:type, true)
# => false
```

spec/hydra/works/services/generic_file/persist_derivatives_output_file_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+describe Hydra::Works::PersistDerivativesOutputFile do
+
+  let(:generic_file)        { Hydra::Works::GenericFile::Base.create }
+  let(:filename)            { "sample-file.pdf" }
+  let(:destination_name)    { File.join(fixture_path, filename) }
+  let(:file)                { File.new(destination_name) }
+  let(:type)                { ::RDF::URI("http://pcdm.org/ExtractedText") }
```
mjgiarlo: Please use `http://pcdm.org/use#ExtractedText` for the URI string.

spec/hydra/works/services/generic_file/persist_derivatives_output_file_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:service_type)        { ::RDF::URI("http://pcdm.org/ServiceFile") }
```
mjgiarlo: Please use `http://pcdm.org/use#ServiceFile` for the URI string.

kevinreiss:
@mjgiarlo @flyingzumwalt Updated per Mike's comments. I shortened the class name and switched over to using #fetch to process the options passed to the service.

kevinreiss:
Removed some cruft.

mjgiarlo:
taking a gander.

mjgiarlo:
:+1: 

Thanks @kevinreiss -- now could you squash your two commits?  then ping me and I will merge.

kevinreiss:
Thanks @mjgiarlo. It should be ready to go now.

mjgiarlo:
OK, @kevinreiss. I'm going to step away for some :tea: and then I'll merge this.

PR #134, opened 2015-06-25
makes versioning optional in AddFile service.

flyingzumwalt:
Also cleans up method signature on AddFile.call()  and UploadFile.call()
refs pulibrary/curation_concerns#24


lib/hydra/works/services/generic_file/add_file.rb
```diff
@@ -44,7 +45,9 @@ def self.call(generic_file, path, type, update_existing=true)
           current_file.save # if we updated an existing file's content we need to save the file explicitly
         end
         generic_file.reload     # this forces the generic_file to see the updated file
-        current_file.create_version # Create version _after_ saving because ActiveFedora::Versionable#create_version tells Fedora to create a version that is a snapshot of the generic_file's current state within Fedora
+        unless versioning == false
```
jcoyne: how about `if versioning` instead?

flyingzumwalt: Good idea.  :updated:

PR #139, opened 2015-07-07
AddFileToGenericFile validates the generic_file before attempting to save

flyingzumwalt:
I'm open to changing how the service responds to this situation, but it needs to do something.  Without the change, it ignores validation completely, meaning that it sometimes attempts to create file versions on unsaved objects and sometimes adds files to fedora even though the parent is not valid (meaning that, for example, a file with a virus would be saved to fedora even if the parent detected the virus)


mjgiarlo:
:+1:

PR #161, opened 2015-07-21
use filters, parent/child API, indexers, deprecate services

elrayle:
Change Log:
* use filter associations
* rework to use parent/child consistently
* add indexers for collections, generic works, generic files
* deprecate atomic service objects
* minimal updates to tests (mostly error messages, removal of unacceptable parent tests, unnecessary save-reload)

### DIFF
--- a/lib/hydra/works/models/concerns/block_child_objects.rb
+++ b/lib/hydra/works/models/concerns/block_child_objects.rb
@@ -1,0 +1,16 @@
+module Hydra::Works
+  # Do not allow aggregation of child objects
+  module BlockChildObjects
+
+    def child_objects= objects
+      raise StandardError, "method `child_objects=' not allowed for #{self}"
+    end
+
+    def child_objects
+      raise StandardError, "method `child_objects' not allowed for #{self}"
+    end
+
+  end
+end
+
+
--- a/lib/hydra/works/models/concerns/collection_behavior.rb
+++ b/lib/hydra/works/models/concerns/collection_behavior.rb
@@ -22,6 +22,33 @@ module Hydra::Works
       type [RDFVocabularies::PCDMTerms.Collection,WorksVocabularies::WorksTerms.Collection]
       include Hydra::Works::AggregatesGenericWorks
       include Hydra::Works::AggregatesCollections
+      include Hydra::Works::BlockChildObjects
+
+      filters_association :members, as: :child_collections, condition: :works_collection?
+      filters_association :members, as: :child_generic_works, condition: :works_generic_work?
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Collection.
+    def works_collection?
+      true
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Generic Work.
+    def works_generic_work?
+      false
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Generic File.
+    def works_generic_file?
+      false
+    end
+
+    def parents
+      aggregated_by
+    end
+
+    def parent_collections
+      aggregated_by.select { |parent| parent.class.included_modules.include?(Hydra::Works::CollectionBehavior) }
     end
 
   end
--- a/lib/hydra/works/models/concerns/generic_file_behavior.rb
+++ b/lib/hydra/works/models/concerns/generic_file_behavior.rb
@@ -20,6 +20,36 @@ module Hydra::Works
       include Hydra::Works::GenericFile::Derivatives
       include Hydra::Works::GenericFile::MimeTypes
       include Hydra::Works::GenericFile::VersionedContent
+      include Hydra::Works::BlockChildObjects
+
+      filters_association :members, as: :child_generic_files, condition: :works_generic_file?
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Collection.
+    def works_collection?
+      false
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Generic Work.
+    def works_generic_work?
+      false
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Generic File.
+    def works_generic_file?
+      true
+    end
+
+    def parents
+      aggregated_by
+    end
+
+    def parent_generic_works
+      aggregated_by.select { |parent| parent.class.included_modules.include?(Hydra::Works::GenericWorkBehavior) }
+    end
+
+    def parent_generic_files
+      aggregated_by.select { |parent| parent.class.included_modules.include?(Hydra::Works::GenericFileBehavior) }
     end
 
   end
--- a/lib/hydra/works/models/concerns/generic_work_behavior.rb
+++ b/lib/hydra/works/models/concerns/generic_work_behavior.rb
@@ -20,10 +20,41 @@ module Hydra::Works
       type [RDFVocabularies::PCDMTerms.Object,WorksVocabularies::WorksTerms.GenericWork]
       include Hydra::Works::AggregatesGenericFiles
       include Hydra::Works::AggregatesGenericWorks
+      include Hydra::Works::BlockChildObjects
+
+      filters_association :members, as: :child_generic_works, condition: :works_generic_work?
+      filters_association :members, as: :child_generic_files, condition: :works_generic_file?
     end
 
     def contains= files
       raise NoMethodError, "works can not directly contain files.  You must add a GenericFile to the work's members and add files to that GenericFile."
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Collection.
+    def works_collection?
+      false
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Generic Work.
+    def works_generic_work?
+      true
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Generic File.
+    def works_generic_file?
+      false
+    end
+
+    def parents
+      aggregated_by
+    end
+
+    def parent_generic_works
+      aggregated_by.select { |parent| parent.class.included_modules.include?(Hydra::Works::GenericWorkBehavior) }
+    end
+
+    def parent_collections
+      aggregated_by.select { |parent| parent.class.included_modules.include?(Hydra::Works::CollectionBehavior) }
     end
 
   end
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -24,6 +24,7 @@ module Hydra
     autoload :AggregatesGenericWorks, 'hydra/works/models/concerns/aggregates_generic_works'
     autoload :AggregatesCollections,  'hydra/works/models/concerns/aggregates_collections'
     autoload :GenericFile,            'hydra/works/models/generic_file'
+    autoload :BlockChildObjects,      'hydra/works/models/concerns/block_child_objects'
 
     # collection services
     autoload :AddCollectionToCollection,           'hydra/works/services/collection/add_collection'
--- a/lib/hydra/works/models/concerns/aggregates_collections.rb
+++ b/lib/hydra/works/models/concerns/aggregates_collections.rb
@@ -2,14 +2,18 @@ module Hydra::Works
   # Allows instances of a class to aggregate (pcdm:hasMember) hydra-works collections
   module AggregatesCollections
 
+    def child_collection_ids
+      child_collections.map(&:id)
+    end
+
     def collections= collections
-      raise ArgumentError, "each collection must be a hydra works collection" unless collections.all? { |c| Hydra::Works.collection? c }
-      raise ArgumentError, "a collection can't be an ancestor of itself" if collection_ancestor?(collections)
-      self.members = self.generic_works + collections
+      warn "[DEPRECATION] `collections=` is deprecated.  Please use `child_collections=` instead.  This has a target date for removal of 07-31-2015"
+      self.child_collections = collections
     end
 
     def collections
-      members.to_a.select { |m| Hydra::Works.collection? m }
+      warn "[DEPRECATION] `collections` is deprecated.  Please use `child_collections` instead.  This has a target date for removal of 07-31-2015"
+      self.child_collections
     end
 
   end
--- a/lib/hydra/works/models/concerns/aggregates_generic_files.rb
+++ b/lib/hydra/works/models/concerns/aggregates_generic_files.rb
@@ -2,18 +2,18 @@ module Hydra::Works
   # Allows instances of a class to aggregate (pcdm:hasMember) hydra-works generic files
   module AggregatesGenericFiles
 
+    def child_generic_file_ids
+      child_generic_files.map(&:id)
+    end
+
     def generic_files= generic_files
-      raise ArgumentError, "each generic_file must be a hydra works generic file" unless generic_files.all? { |w| Hydra::Works.generic_file? w }
-      raise ArgumentError, "a generic file can't be an ancestor of itself" if object_ancestor?(generic_files)
-      if self.respond_to?(:generic_works)
-        self.members = self.generic_works + generic_files
-      else
-        self.members = generic_files
-      end
+      warn "[DEPRECATION] `generic_files=` is deprecated.  Please use `child_generic_files=` instead.  This has a target date for removal of 07-31-2015"
+      self.child_generic_files = generic_files
     end
 
     def generic_files
-      members.to_a.select { |m| Hydra::Works.generic_file? m }
+      warn "[DEPRECATION] `generic_files` is deprecated.  Please use `child_generic_files` instead.  This has a target date for removal of 07-31-2015"
+      self.child_generic_files
     end
   end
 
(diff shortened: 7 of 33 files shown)


lib/hydra/works/models/concerns/block_child_objects.rb
```diff
@@ -0,0 +1,21 @@
+module Hydra::Works
+  # Do not allow aggregation of child objects
+  module BlockChildObjects
+
+    def child_objects= objects
+      raise NoMethodError, "undefined method `child_objects=' for #{self}"
```
jcoyne: This is kind of a lie isn't it? Can we raise something besides NoMethodError?

lib/hydra/works/models/concerns/block_child_objects.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def child_objects
+      raise NoMethodError, "undefined method `child_objects' for #{self}"
```
jcoyne: Raise something else?

lib/hydra/works/models/concerns/block_child_objects.rb
```diff
@@ -0,0 +1,21 @@
...
+    end
+
+    def respond_to? method
```
jcoyne: I don't think this is a good idea.  Why do we need to lie about what the object actually responds to?

lib/hydra/works/models/concerns/collection_behavior.rb
```diff
@@ -22,6 +22,39 @@ module CollectionBehavior
...
+    end
+
+    module ClassMethods
+      def indexer
+        Hydra::Works::CollectionIndexer
+      end
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Collection.
+    def works_collection?
```
jcoyne: can we just call this `collection?`

lib/hydra/works/models/concerns/collection_behavior.rb
```diff
@@ -22,6 +22,39 @@ module CollectionBehavior
...
+      true
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Generic Work.
+    def works_generic_work?
```
jcoyne: can we just call this `generic_work?`

lib/hydra/works/models/concerns/collection_behavior.rb
```diff
@@ -22,6 +22,39 @@ module CollectionBehavior
...
+      false
+    end
+
+    # @return [Boolean] whether this instance is a Hydra::Works Generic File.
+    def works_generic_file?
```
jcoyne: When would this method be called?

lib/hydra/works/models/concerns/collection_behavior.rb
```diff
@@ -22,6 +22,39 @@ module CollectionBehavior
...
+      false
+    end
+
+    def parents
+      aggregated_by
+    end
+
+    def parent_collections
+      aggregated_by.select { |parent| parent.class.included_modules.include?(Hydra::Works::CollectionBehavior) }
```
jcoyne: could this be  `parent.try(:collection?)`

lib/hydra/works/models/concerns/generic_file_behavior.rb
```diff
@@ -20,6 +20,42 @@ module GenericFileBehavior
       include Hydra::Works::GenericFile::Derivatives
       include Hydra::Works::GenericFile::MimeTypes
       include Hydra::Works::GenericFile::VersionedContent
+      include Hydra::Works::BlockChildObjects
+
+      filters_association :members, as: :child_generic_files, condition: :works_generic_file?
```
jcoyne: We would never have parent_generic_files, so can this just be `generic_files`?

lib/hydra/works/models/concerns/generic_file_behavior.rb
```diff
@@ -20,6 +20,42 @@ module GenericFileBehavior
...
+    def works_generic_file?
+      true
+    end
+
+    def parents
+      aggregated_by
+    end
+
+    def parent_generic_works
+      aggregated_by.select { |parent| parent.class.included_modules.include?(Hydra::Works::GenericWorkBehavior) }
```
jcoyne: could this be parent.try(:generic_work?)

lib/hydra/works/models/concerns/generic_file_behavior.rb
```diff
@@ -20,6 +20,42 @@ module GenericFileBehavior
...
+    end
+
+    def parent_generic_files
+      aggregated_by.select { |parent| parent.class.included_modules.include?(Hydra::Works::GenericFileBehavior) }
```
jcoyne: Does this make sense? I thought GenericFiles were always leaf nodes.

lib/hydra/works/models/concerns/generic_work_behavior.rb
```diff
@@ -20,11 +20,48 @@ module GenericWorkBehavior
       type [RDFVocabularies::PCDMTerms.Object,WorksVocabularies::WorksTerms.GenericWork]
       include Hydra::Works::AggregatesGenericFiles
       include Hydra::Works::AggregatesGenericWorks
+      include Hydra::Works::BlockChildObjects
+
+      filters_association :members, as: :child_generic_works, condition: :works_generic_work?
+      filters_association :members, as: :child_generic_files, condition: :works_generic_file?
```
jcoyne: all `GenericFiles` are children, so can we just say `generic_files`?

lib/hydra/works/models/concerns/generic_work_behavior.rb
```diff
@@ -20,11 +20,48 @@ module GenericWorkBehavior
       type [RDFVocabularies::PCDMTerms.Object,WorksVocabularies::WorksTerms.GenericWork]
       include Hydra::Works::AggregatesGenericFiles
       include Hydra::Works::AggregatesGenericWorks
+      include Hydra::Works::BlockChildObjects
+
+      filters_association :members, as: :child_generic_works, condition: :works_generic_work?
```
jcoyne: While a `PCDM::Work` could be a parent, I don't think a GenericWork could ever be a parent of a GenericWork, could it?

lib/hydra/works/models/concerns/block_child_objects.rb
elrayle: Yep, it is a sort of lie.  Do you have a suggestion of an alternate error to raise?

lib/hydra/works/models/concerns/collection_behavior.rb
elrayle: The way the code is written, there is a difference between a pcdm:collection and a works:collection.  A pcdm:collection aggregates pcdm:objects.  A works:collection aggregates works:collections and works:generic_works.

On an intuitive level, I am not convinced that this is exactly what we want, but haven't had time to work through examples of collections that mix use of pcdm:objects and works:generic_works.

lib/hydra/works/models/concerns/collection_behavior.rb
elrayle: I was following the pattern set at pcdm using pcdm_collection? and pcdm_object.  I do not have a problem using just generic_work?  It seems intuitive enough.  I would like to see consistency.  What happens if sufia defines a generic_work?  In the current pattern, there would be works_generic_work?  and sufia_generic_work?

lib/hydra/works/models/concerns/collection_behavior.rb
elrayle:
```ruby
generic_work.child_generic_files += [col1]
```
This will call col1.works_generic_file? and return false which will cause the filter association code to print out a nice message for the caller telling them they passed col1 which is the wrong type.

lib/hydra/works/models/concerns/collection_behavior.rb
elrayle: This isn't new code.  It is copied from pcdm.  If this should be done differently, I prefer a new issue be opened to address that.

lib/hydra/works/models/concerns/generic_file_behavior.rb
elrayle: Why couldn't generic_files contain generic_files?  I have a use case for this.

lib/hydra/works/models/concerns/generic_file_behavior.rb
elrayle: Same comment as above... new issue.

lib/hydra/works/models/concerns/block_child_objects.rb
jcoyne: How about just StandardError "not allowed to call child_objects= on #{self.class}"

lib/hydra/works/models/concerns/collection_behavior.rb
jcoyne: But doesn't the members association (which the filter delegates to) responsible for validating that?

lib/hydra/works/models/concerns/collection_behavior.rb
jcoyne: But it *is* new code in this gem.

lib/hydra/works/models/concerns/generic_file_behavior.rb
jcoyne: If a GenericFile is supposed to represent a file like a bitstream object, how could it contain more files?

lib/hydra/works/models/concerns/generic_file_behavior.rb
jcoyne: I disagree. Putting it in this gem makes it new.

lib/hydra/works/models/concerns/generic_work_behavior.rb
elrayle: There is no PCDM::Work.  Is this a typo?

There are lots of use cases for GenericWork as a child of another GenericWork. 

Ex.  https://docs.google.com/document/d/1FxYyhVkjIuvOq1d-MS2dzKTElIfNSk01ytwiAB5DZiw/edit

terrellt: :+1: to generic works being parents of generic works.

jcoyne: Okay. I thought GenericWork was a restricted case where it didn't have work parents. (@elrayle yes, I meant PCDM::Object)

lib/hydra/works/models/concerns/collection_behavior.rb
jcoyne: I would think you'd want to use one method name, and each override could narrow/change the scope as appropriate. If sufia defines sufia_generic_work? then this filter won't use it.

lib/hydra/works/models/concerns/generic_file_behavior.rb
elrayle: I can see your point that it is new to Hydra:Works.  If you want to go the route of doing this different from what is at PCDM, then I will extract out all the parent code in Hydra:Works and open a separate issue fixing pcdm parents and adding parents in Hydra:Works.  I believe it can be removed without breaking other parts.

lib/hydra/works/models/concerns/collection_behavior.rb
elrayle: It is not 'new' in the sense that it is refactoring code being ported to Hydra:Works from PCDM.  This entire PR is adding no 'new' code, but is making Hydra:Works consistent with PCDM.  Any issue you raise here applies to PCDM as it exists at master today.

lib/hydra/works/models/concerns/collection_behavior.rb
elrayle: Yes, and col1 would still need to respond to `works_generic_file?`

```ruby
generic_work.members += [col1]
```
should call col1.works_generic_file? and col1.works_generic_work?  I'd have to confirm that this is actually happening, but it is what I expect.

lib/hydra/works/models/concerns/collection_behavior.rb
elrayle: This is the style of validation that @terrellt put in that was accepted.  I am following the previously established pattern, not creating a new one.

lib/hydra/works/models/concerns/collection_behavior.rb
jcoyne: Okay, this isn't even important. I just thought it more readable.

lib/hydra/works/models/concerns/generic_file_behavior.rb
jcoyne: I don't really care. I just thought it's easier to read and leverages the other code you're adding in this PR

jcoyne:
Why are we adding indexers here? These are very slow and I'm not sure we should make the behavior slow by default.  Can you say more about the motivation?

lib/hydra/works/models/concerns/collection_behavior.rb
jcoyne: shouldn't we treat this case just like any other object that doesn't respond to that method?  e.g. (`generic_work.child_generic_files += [Foo.new]`) Why do we have a special case for putting Collections into the generic_file association?  You should never do that, just like you should never put a `Foo` on there?

elrayle: This is the validation approach brought in by @terrellt that replaced the approach used previously.  See https://github.com/projecthydra-labs/hydra-pcdm/pull/124/files and https://github.com/projecthydra-labs/hydra-pcdm/issues/151.  If you don't agree with the approach brought in via this PR, I recommend accepting this PR as is and adding an issue at pcdm and works describing the alternate approach.

elrayle:
I can remove the indexers.  I was under the impression they were needed and added them to be consistent with pcdm.  For my edification, are the indexers brought in by the end applications instead?  Otherwise, how are generic_works and generic_files indexed?

jcoyne:
I'm for simplicity, so I don't want to put in the indexers unless we need to. Currently curation concerns has some indexers because it needs certain fields to be indexed. It's always easier to add code later than to take it out.

lib/hydra/works/models/concerns/generic_file_behavior.rb
elrayle: Opened an issue to discuss this further.  The PR does not change the behaviors relative to the current code at master.

https://github.com/projecthydra-labs/hydra-works/issues/165

elrayle:
removed indexers

lib/hydra/works/models/concerns/collection_behavior.rb
jcoyne: It's already changed in PCDM: https://github.com/projecthydra-labs/hydra-pcdm/pull/155/files Can we follow the same pattern here?

escowles:
I'm seeing an error when I try to upgrade curation_concerns to this hydra-works at this commit or later:

https://gist.github.com/escowles/7d98677d4304e9ead46a

I don't see any code calling child_objects, instead I see a test trying to create a Collection object:

https://github.com/escowles/curation_concerns/blob/master/spec/abilities/collection_abilities_spec.rb#L12

which uses a factory:

https://github.com/escowles/curation_concerns/blob/master/spec/factories/collections.rb

The stacktrace ends with a module which calls to_solr.super:

https://github.com/escowles/curation_concerns/blob/master/curation_concerns-models/app/models/concerns/curation_concerns/human_readable_type.rb#L15

None of this code is explicitly calling child_objects.  I don't see any mention of child_objects in curation_concerns or hydra-works (except in block_child_objects.rb).

tampakis:
I'm seeing this too.

terrellt:
Further stack trace:
```
Exception: StandardError: method `child_objects' not allowed for Test collection title
--
0: /Users/terrellt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/hydra-works-a5684fe4a525/lib/hydra/works/models/concerns/block_child_objects.rb:10:in `child_objects'
1: /Users/terrellt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/hydra-pcdm-6e5d76306de3/lib/hydra/pcdm/models/concerns/child_objects.rb:4:in `child_object_ids'
2: /Users/terrellt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/hydra-pcdm-6e5d76306de3/lib/hydra/pcdm/object_indexer.rb:5:in `block in generate_solr_document'
3: /Users/terrellt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/hydra-pcdm-6e5d76306de3/lib/hydra/pcdm/object_indexer.rb:4:in `tap'
4: /Users/terrellt/.rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/bundler/gems/hydra-pcdm-6e5d76306de3/lib/hydra/pcdm/object_indexer.rb:4:in `generate_solr_document'
```

terrellt:
So what's happening is it's trying to index child object IDs to solr, and it's blocked. A custom indexer would probably help.

terrellt:
We probably didn't catch this because we removed the indexer tests at the work level.

tampakis:
Yeah I came across that as well. This error comes up if you try calling `work.generic_files.size`

tampakis:
my bad, that's something else that I came across

terrellt:
Just as a note here for future parties, these problems are dealt with now.

PR #166, opened 2015-07-23
Update activefedora-aggregation version to ~> 0.3

grosscol:
Also, update gitignore for config dir and vim swp files.  We use a different config/fedora.yml than the default and others might do so as well.  It's not required for hydra-works so config/ can be ignored for the time being.

### DIFF
--- a/hydra-works.gemspec
+++ b/hydra-works.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "hydra-pcdm", "~> 0.0.1"
   spec.add_dependency "hydra-derivatives", "~> 1.1.0"
-  spec.add_dependency "activefedora-aggregation", "~> 0.2.0"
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -8,12 +9,17 @@
 /spec/reports/
 spec/internal
 /tmp/
+/config
```
jcoyne: Please put this in `.git/info/exclude` not .gitignore.

.gitignore
```diff
@@ -8,12 +9,17 @@
...
+
+#Ignore config and intermediate file extensions
 *.bundle
 *.so
 *.o
 *.a
+*.swp
```
jcoyne: Put this in your global git ignore. See https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

grosscol:
Jose pinged me on this one.  Changes made.  @jcoyne Thanks for the git directions.

PR #175, opened 2015-07-28
Add indexer for child generic works

### LINKED ISSUE #109: BUG: solrize fails with "does not respond to `include?`"
In add_collection_spec, test 'should solrize member ids' fails
```
expect(subject.to_solr["generic_works_ssim"]).to include(generic_work1.id,generic_work2.id)
expect(subject.to_solr["generic_works_ssim"]).not_to include(collection2.id,collection1.id,collection3.id)
expect(subject.to_solr["collections_ssim"]).to include(collection2.id,collection1.id,collection3.id)
expect(subject.to_solr["collections_ssim"]).not_to include(generic_work1.id,generic_work2.id)
```
This code was copied from pcdm tests.  Only the test was copied.  It could be that some additional setup is required for to_solr method to work properly.

Similar tests fail in each of the Add services.

terrellt:
Closes #109

### DIFF
--- a/lib/hydra/works/indexers/collection_indexer.rb
+++ b/lib/hydra/works/indexers/collection_indexer.rb
@@ -1,0 +1,9 @@
+module Hydra::Works
+  class CollectionIndexer < Hydra::PCDM::CollectionIndexer
+    def generate_solr_document
+      super.tap do |solr_doc|
+        solr_doc["child_generic_work_ids_ssim"] = object.child_generic_work_ids
+      end
+    end
+  end
+end
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -53,6 +53,9 @@ module Hydra
     autoload_under 'services/generic_file' do
       autoload :AddFileToGenericFile
     end
+    autoload_under 'indexers' do
+      autoload :CollectionIndexer
+    end
     autoload :AddGenericFileToGenericFile,         'hydra/works/services/generic_file/add_generic_file'
     autoload :AddRelatedObjectToGenericFile,       'hydra/works/services/generic_file/add_related_object'
     autoload :GetGenericWorksFromGenericFile,      'hydra/works/services/generic_file/get_generic_works'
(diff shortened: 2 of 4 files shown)


lib/hydra/works/indexers/collection_indexer.rb
```diff
@@ -0,0 +1,9 @@
+module Hydra::Works
+  class CollectionIndexer < Hydra::PCDM::CollectionIndexer
```
escowles: Since GenericWorks also have child GenericWorks, maybe this should be called ChildGenericWorksIndexer?

terrellt: Unfortunately I don't think we can do that, because of the inheritance. Wouldn't GenericWorks need the GenericObjectIndexer from PCDM?

We probably want to rearchitect the indexers to avoid duplication via composition, but I'd have to start that lower down the stack.

jcoyne:
Why does Hydra-Works need to index it's child generic works? What's the real use case? I think CurationConcerns already does this and it'd be extra slow to have it happen more than once.

lib/hydra/works/indexers/collection_indexer.rb
escowles: That makes sense -- though I'm not sure we need to start the composition lower in the stack.  Could you pull out the generate_solr_document body, and let CollectionIndexer and GenericWork indexer call it?

escowles:
None of the specs say anything about how it would be used -- they just check the solr_doc for the ids.

Unless it's needed for the characterization, derivatives, etc., I'm fine with just getting rid of it.

escowles:
Related, there are a total of 7 tests that use to_solr -- and all of them are skipped because of #109.  There are no uses of to_solr in the main code at all.  So I'm for just removing those tests.

terrellt:
I'd expect it to be needed at the Sufia level, but if Curation Concerns can handle it, then I'm alright with it.

escowles:
I've opened #177 to remove the solr tests.

PR #177, opened 2015-07-28
Removing solr tests

escowles:
See #175


mjgiarlo:
@terrellt OK to merge this?

terrellt:
I think I'm :+1:. We can always go back on the decision later.

Issue #94, opened 2015-05-13
Move Virus Checking functionality from Sufia to Hydra::Works


flyingzumwalt:
Move Sufia::GenericFile::VirusCheck to Hydra::Works::File::VirusCheck

**Note**: Provide this in hydra-works but make it optional -- don't include in FileBehaviors

escowles:
The other half of this (making curation_concerns use the updated hydra-works) is tracked in https://github.com/projecthydra-labs/curation_concerns/issues/37

PR #178, opened 2015-07-30
Migrating virus check functionality from CurationConcerns

escowles:
Addresses #94 (part 1 of 2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,22 @@ f1.content = "The quick brown fox jumped over the lazy dog."
 bf1.save
 ```
 
+## Virus Detection
+
+To turn on virus detection, install clamav on your system and add the `clamav` gem to your Gemfile
+
+    gem 'clamav'
+
+Then include the `VirusCheck` module in your `GenericFile` class:
+
+```ruby
+class BookFiles < ActiveFedora::Base
+  include Hydra::Works::GenericFileBehavior
+  include Hydra::Works::GenericFile::VirusCheck
+end
+```
+
+
 ## Access controls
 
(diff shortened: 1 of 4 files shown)


jcoyne:
This seems like business logic more than modeling. As such does it belong in hydra-works?  I'm not against it being here, I'm just making sure we're considering it thoughtfully.

escowles:
@jcoyne I think the reasoning for moving this down to hydra-works is to have all of the core functionality here (generating derivatives, creating versions, checking for viruses, etc.) so you don't have to use curation_concerns to get that functionality.

jcoyne:
Yes, I now see that derivatives are here, so it probably belongs with that.

jcoyne:
Should there be something added to the Readme saying "Add clamav to your gemfile to enable virus checking"

jpstroop:
Close projecthydra-labs/curation_concerns#37 and move [this part of CurationConcern's README](https://github.com/projecthydra-labs/curation_concerns#virus-detection) here when this goes through.

mjgiarlo:
@escowles Do you want to include the README change that @jpstroop suggests in this PR, or attach that to part 2?

:+1:

escowles:
@mjgiarlo I'll add the README update to this PR -- and then part 2 will be updating curation_concerns to use this virus check instead of its own

jpstroop:
Actually, it's even more than that; [virus check is also in CurationConcerns](https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/models/concerns/curation_concerns/generic_file/virus_check.rb) :cry: , so that should be removed. I'll update projecthydra-labs/curation_concerns#37

jpstroop:
Doh!

mjgiarlo:
@escowles OK. I'll merge this PR as soon as you tweak the README. (I won't bother waiting on travis for a README change -- we already know the suite passes.) Ping me when that's ready.

Issue #164, opened 2015-07-21
 deprecate old style validation


elrayle:
deprecate validations in [works.rb](https://github.com/projecthydra-labs/hydra-works/blob/master/lib/hydra/works.rb) and use works_generic_file? works_generic_work? and works_collection? instead.

```ruby
###  DEPRECATE each of the following
    # model validations
    def self.collection? collection
      return false unless collection.respond_to? :type
      collection.type.include? WorksVocabularies::WorksTerms.Collection
    end

    def self.generic_work? generic_work
      return false unless generic_work.respond_to? :type
      generic_work.type.include? WorksVocabularies::WorksTerms.GenericWork
    end

    def self.generic_file? generic_file
      return false unless generic_file.respond_to? :type
      generic_file.type.include? WorksVocabularies::WorksTerms.GenericFile
    end
```

PR #182, opened 2015-07-31
deprecate methods in works.rb for issue 164

blancoj:
deprecate validations in works.rb and use works_generic_file? works_generic_work? and works_collection? instead.

This is to address this issue:  https://github.com/projecthydra-labs/hydra-works/issues/164

If this looks good, I can fix spec/hydra/works_spec.rb to use the new methods and not the deprecated methods.  Should I do that as part of this issue?

-Jose

### DIFF
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -64,20 +64,20 @@ module Hydra
     autoload :GenerateThumbnail,                   'hydra/works/services/generic_file/generate/thumbnail'
     autoload :PersistDerivative,                   'hydra/works/services/generic_file/persist_derivative'
 
-    # model validations
+    # model validations - these have been deprecated
     def self.collection? collection
-      return false unless collection.respond_to? :type
-      collection.type.include? WorksVocabularies::WorksTerms.Collection
+      warn "[DEPRECATION] `collection? collection` is deprecated.  Please use `works_collection?` instead.  This has a target date for removal of 07-31-2015"
+      collection.works_collection?
     end
 
     def self.generic_work? generic_work
-      return false unless generic_work.respond_to? :type
-      generic_work.type.include? WorksVocabularies::WorksTerms.GenericWork
+      warn "[DEPRECATION] `generic_work? generic_work` is deprecated.  Please use `works_generic_work?` instead.  This has a target date for removal of 07-31-2015"
+      generic_work.works_generic_work?
     end
 
     def self.generic_file? generic_file
-      return false unless generic_file.respond_to? :type
-      generic_file.type.include? WorksVocabularies::WorksTerms.GenericFile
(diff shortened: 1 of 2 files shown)


lib/hydra/works.rb
```diff
@@ -64,21 +64,38 @@ module Works
     autoload :GenerateThumbnail,                   'hydra/works/services/generic_file/generate/thumbnail'
     autoload :PersistDerivative,                   'hydra/works/services/generic_file/persist_derivative'
 
-    # model validations
-    def self.collection? collection
+    # model validations 
+    def self.works_collection? collection
```
escowles: These methods should be removed -- these same methods are implemented in models/concerns/work_behavior.rb, e.g.: [#works_collection](https://github.com/blancoj/hydra-works/blob/issue_164/lib/hydra/works/models/concerns/generic_work_behavior.rb#L34)

blancoj: Esme,

Once I remove these methods from hydra/works/works.rb

works_collection?
works_generic_file?
works_generic_work?

because they live in hydra/works/model/concerns/generic_file_behavior.rb

What do I do about spec/hydra/work_spec.rb  ?

The test that ask for

works_collection?
works_generic_file?
works_generic_work?

come back as undefined.

escowles: The test should be able to check

```ruby
works_gwork.works_generic_work?
```

instead of

```ruby
Hydra::Works.works_generic_work? works_gwork 
```

Does that work?

blancoj: Yes that works, but what about these types of tests:

old way:
        expect( Hydra::Works.works_collection? pcdm_coll ).to be false

new way?:
        expect( pcdm_coll.works_collection? ).to be false

both of these fail. Should I get rid of these tests?

escowles: I would expect that test:

```ruby
expect( pcdm_coll.works_collection? ).to be false
```

to raise a NoMethodError -- so maybe the more appropriate test for non-Hydra::Works objects would be something like:

```ruby
expect( pcdm_coll.respond_to? :works_collection ).to be false
```

mjgiarlo: Bonus points for using a more specific matcher:

```ruby
expect(pcdm_col1).not_to respond_to(:works_collection)
```

blancoj: Yes, it does return no method, so I will change it to be like you propose,
and you can let me know what you think when I push the PR.

Thanks!
Jose

escowles:
The Travis build failed, and I'm seeing the same failures locally -- have you run the full test suite?  It looks like those failures are related to the methods being used in the remaining services.

blancoj:
No I have not.  Let me try that and see what I find out.

PR #183, opened 2015-08-03
Test coverage for Hydra::Works::GenericFile::Derivatives.

### LINKED ISSUE #152: Hydra::Works Derivatives Generation has test coverage and relies on correct output_file_service
This will be complete when: 
* hydra-works configures `Hydra::Derivatives.output_file_service` to use the PersistDerivative Service from #132 
* hydra-works has test coverage for `Hydra::Works::GenericFile::Derivatives`

Related: hydra-derivatives should have test coverage to ensure that Processor.new will default to using the output_file_service you set in Hydra::Derivatives.output_file_service: projecthydra/hydra-derivatives#65

tampakis:
Closes #152.
 - Included all [sufia-models derivatives](https://github.com/projecthydra/sufia/blob/master/sufia-models/app/models/concerns/sufia/generic_file/derivatives.rb#L9-L22) in GenericFile::Derivatives.
 - GenericFile now [directly contains](https://github.com/projecthydra-labs/hydra-works/compare/derivatives_spec_refactor?expand=1#diff-61227a90d08989711495f539862d4e7dR13) all of the derivatives defined for GenericFile.
 - Update [readme](https://github.com/projecthydra-labs/hydra-works/compare/derivatives_spec_refactor?expand=1#diff-04c6e90faac2675aa89e2176d2eec7d8R62) to show how to configure additional derivatives.
 - Skip [ffmpeg-related specs](https://github.com/projecthydra-labs/hydra-works/compare/derivatives_spec_refactor?expand=1#diff-8b0e62a1ec782e7f1facccb31b5641caR120) and [sofice specs](https://github.com/projecthydra-labs/hydra-works/compare/derivatives_spec_refactor?expand=1#diff-8b0e62a1ec782e7f1facccb31b5641caR86) for Travis because they are not configured. Is this something we should consider supporting in CI?
 - Ensure that files are [readable](https://github.com/projecthydra-labs/hydra-works/compare/derivatives_spec_refactor?expand=1#diff-98db26374e8da5b14c20a72e615ed2e9R26). Some of the processors in H::Derivatives pass raw binary strings instead of IO objects.
 - Re-add mime_type as an option for [AddFileToGenericFileService](https://github.com/projecthydra-labs/hydra-works/compare/derivatives_spec_refactor?expand=1#diff-337e76c89380858159ef5281de6f183cR11) because IO objects don't respond to .mime_type. All H:Derivative processors pass the mime_type option to the output_file_service.


grosscol:
Instead of adding mime_type to the signature to accommodate the HydraDerivatives use, we're going to refactor HydraDerivatives to wrap the IO objects in a class that has their attributes.  This will be similar to how CurationConcerns operates.

README.md
```diff
@@ -59,6 +59,70 @@ f1.content = "The quick brown fox jumped over the lazy dog."
 bf1.save
 ```
 
+## Extending Directly Contained Derivatives
```
mjgiarlo: For simplicity's sake, can we name this heading something like "Custom Derivatives"? While I appreciate the specificity around how the derivs will be contained (in LDP parlance), I suspect many users of Hydra::Works may not have much knowledge of the intricacies of LDP.

README.md
```diff
@@ -59,6 +59,70 @@ f1.content = "The quick brown fox jumped over the lazy dog."
...
+
+Hydra Works comes with the ability to attach [Hydra Derivatives](https://github.com/projecthydra/hydra-derivatives) to a GenericFile. This allows a GenericFile of an image, video, audio file, or document type to directly contain appropriate derivative files such as thumbnail, extracted_text, mp3, ogg, mp4, and webm. The derivatives are configured in `lib/hydra/works/models/concerns/generic_file/contained_files.rb` and `lib/hydra/works/models/concerns/generic_file/derivatives.rb`.
```
mjgiarlo: What does "appropriate" mean here?  This could confuse people, so it may be easier to `s/appropriate //`

README.md
```diff
@@ -59,6 +59,70 @@ f1.content = "The quick brown fox jumped over the lazy dog."
...
+
+More directly contained derivatives can be configured for the Persist Derivatives service by extending `contained_files` and `makes_derivatives`, as long as the processor is defined in Hydra Derivatives.
```
mjgiarlo: In line with my comments above, `s/directly contained //`

mjgiarlo: I'd also strike the reference to the `Persist Derivatives` service, and say something like this here:

```More derivatives can be generated by extending `contained_files` and `makes_derivatives`, as long as  processors for them are defined in Hydra::Derivatives (linked above).```

File: README.md [hunk 4]
```diff
@@ -59,6 +59,70 @@ f1.content = "The quick brown fox jumped over the lazy dog."
...
+
+### contained_files
+```
+# lib/hydra/works/models/concerns/generic_file/contained_files.rb
```
mjgiarlo: So the mechanism for adding custom derivatives is monkeypatching Hydra::Works modules in `lib/`?  I'm -0 on that pattern here, but I don't care enough about this to hold you up unless others also care.  I'd prefer to see Hydra::Works generate an initializer into an app with some commented-out code that can be used to customize things. Something like the following:

```ruby
# config/initializers/hydra_works.rb
Hydra::Works.configure
  # add_derivative(name: :new_derivative_name, uri: 'http://pcdm.org/use#ServiceFile', format: 'png', size: '400x400', thumbnail: { format: 'jpg', size: '338x493' }, mime_types: ['application/pdf'])
end
```

The cost of this approach is that it's going to require more code.  The benefits are providing a less troublesome upgrade path (if I have to copy/paste code into a monkeypatch, and the original code in Hydra::Works evolves, I need to track that manually) and abstracting away from users the details of how derivatives are generated (i.e., knowledge of `contained_files` and `makes _derivatives`).

I also don't love the API I sketched out above, so I'm happy for others to chime in. @terrellt @jcoyne @escowles @grosscol @elrayle @jpstroop etc.

lib/hydra/works/models/concerns/generic_file/contained_files.rb
```diff
@@ -10,6 +10,9 @@ module Hydra::Works::GenericFile::ContainedFiles
     directly_contains_one :original_file, through: :files, type: ::RDF::URI("http://pcdm.org/use#OriginalFile"), class_name: "Hydra::PCDM::File"
     directly_contains_one :thumbnail, through: :files, type: ::RDF::URI("http://pcdm.org/use#ThumbnailImage"), class_name: "Hydra::PCDM::File"
     directly_contains_one :extracted_text, through: :files, type: ::RDF::URI("http://pcdm.org/use#ExtractedText"), class_name: "Hydra::PCDM::File"
+    directly_contains_one :mp3, through: :files, type: ::RDF::URI("http://pcdm.org/use#ServiceFile"), class_name: "Hydra::PCDM::File"
```
mjgiarlo: Can you talk about why we're changing the default derivatives to include these four new ones?  I ask because of the comment on line 4 -- I'm not sure if it's meant to be normative or descriptive.

lib/hydra/works/models/concerns/generic_file/derivatives.rb
```diff
@@ -5,24 +5,27 @@ module Derivatives
...
-          obj.transform_file :original_file, { thumbnail: { format: 'jpg', size: '200x150>', datastream: 'thumbnail' } }, processor: :work
+          obj.transform_file :original_file, thumbnail: { format: 'jpg', size: '200x150>' }
         end
+        
       end
 
     end
 
-    # This bit was not in Sufia, but needs to be here for derivatives to work
-    # TODO: delegate mime type somewhere else?
```
mjgiarlo: Does this overridden method still need to be here?  (Will the work @grosscol is doing in hydra-derivatives obviate the need for this method override?)

lib/hydra/works/services/generic_file/add_file_to_generic_file.rb
```diff
@@ -8,7 +8,7 @@ class AddFileToGenericFile
     # @param [Boolean] update_existing whether to update an existing file if there is one. When set to true, performs a create_or_update. When set to false, always creates a new file within generic_file.files.
     # @param [Boolean] versioning whether to create new version entries (only applicable if +type+ corresponds to a versionable file)
 
-    def self.call(generic_file, file, type, update_existing: true, versioning: true)
+    def self.call(generic_file, file, type, update_existing: true, versioning: true, mime_type: nil)
```
mjgiarlo: Just a note that the MIME type-related changes to this file are the ones @grosscol is working on in hydra-derivatives, so this is slated to go away once that's done.

README.md
grosscol: Maybe s/appropriate/associated/

lib/hydra/works/services/generic_file/persist_derivative.rb
```diff
@@ -9,17 +9,29 @@ class PersistDerivative < Hydra::Derivatives::PersistOutputFileService
...
+      type = destination_name.gsub(/^original_file_/, '').to_sym
       update_existing = opts.fetch(:update_existing, true)
       versioning = opts.fetch(:versioning, true)
-      Hydra::Works::AddFileToGenericFile.call(object, file, type, update_existing: update_existing, versioning: versioning)
+
+      # copy original_file stream to audio derivative if original_file is .ogg or .mp3
+      file = object.original_file.content if self.audio_format_match?(object.original_file.mime_type, type)
+
+      # file must respond to #read
+      readable = file.class == String ? StringIO.new(file) : file
```
mjgiarlo: This is also part of the code that will go away when @grosscol is done, right?

lib/hydra/works/services/generic_file/persist_derivative.rb
```diff
@@ -9,17 +9,29 @@ class PersistDerivative < Hydra::Derivatives::PersistOutputFileService
...
-    def self.call(object, file, opts={})
-      type = opts.fetch(:type, ::RDF::URI("http://pcdm.org/use#ServiceFile"))
+    def self.call(object, file, destination_name, opts={})
+      type = destination_name.gsub(/^original_file_/, '').to_sym
       update_existing = opts.fetch(:update_existing, true)
       versioning = opts.fetch(:versioning, true)
-      Hydra::Works::AddFileToGenericFile.call(object, file, type, update_existing: update_existing, versioning: versioning)
+
+      # copy original_file stream to audio derivative if original_file is .ogg or .mp3
+      file = object.original_file.content if self.audio_format_match?(object.original_file.mime_type, type)
```
mjgiarlo: Can you explain why this special behavior is needed here?

README.md
tampakis: :+1:

lib/hydra/works/services/generic_file/persist_derivative.rb
tampakis: Yeah.

spec/hydra/works/services/persist_derivatives_spec.rb
```diff
@@ -0,0 +1,158 @@
+require 'spec_helper'
```
mjgiarlo: :clap:

README.md
terrellt: So I'm not sure either is necessary - works classes are meant to be inherited, no? So really I'd expect something like...

```ruby
class GenericFile
  include Hydra::Works::GenericWorkBehavior
  directly_contains_one :new_derivative_name, through: :files, type: ::RDF::URI("http://pcdm.org/use#ServiceFile"), class_name: "Hydra::PCDM::File"
  makes_derivatives do |obj|
    case obj.mime_type

      # create :new_derivative_name for pdf in additon to :thumbnail derivative
    when *pdf_mime_types
      obj.transform_file :original_file, { new_derivative_name: { format: 'png', size: '400x400' }, thumbnail: { format: 'jpg', size: '338x493' } }
    end
  end
end
```

I don't think this will work by the way - where's use#ServiceFile come from?

spec/hydra/works/services/generic_file/persist_derivative_spec.rb
```diff
@@ -1,21 +0,0 @@
...
-
-describe Hydra::Works::PersistDerivative do
-
-  let(:generic_file)        { Hydra::Works::GenericFile::Base.new }
-  let(:filename)            { "sample-file.pdf" }
-  let(:file)                { File.open(File.join(fixture_path, filename)) }
-  let(:type)                { ::RDF::URI("http://pcdm.org/use#ExtractedText") }
-  let(:service_type)        { ::RDF::URI("http://pcdm.org/use#ServiceFile") }
-
-  it "persists a service file by default" do
```
mjgiarlo: Can you explain why these tests are going away? Are they already covered by more specific tests below? The level of abstraction seems to be different in this spec.

README.md
mjgiarlo: :ok:

lib/hydra/works/models/concerns/generic_file/contained_files.rb
tampakis: Sufia includes these 4 derivatives so I thought it would be good to pull down that behavior. The `directly_contains_one` asssociation allows you to call derivatives directly on the generic_file as follows: `generic_file.mp3`
https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/sufia-models/app/models/concerns/sufia/generic_file/derivatives.rb

lib/hydra/works/services/generic_file/persist_derivative.rb
mjgiarlo: :+1:

lib/hydra/works/models/concerns/generic_file/contained_files.rb
grosscol: Seems like the ogg, mp3, webm, and mp4 derivatives might be better and an example of a application specific Concern name something like "ContainedFilesExtras" or "EvenMoreContainedFiles".

I like them as examples of common use cases for extending the contained files.

README.md
jpstroop: ServiceFile is in the use ontology:

mjgiarlo: It's from the PCDM use extensions vocab. https://github.com/duraspace/pcdm/blob/master/pcdm-ext/use.rdf

terrellt: I guess I meant where is it assigned, but transform_file is probably using the accessor which will auto-add the type.

lib/hydra/works/models/concerns/generic_file/derivatives.rb
grosscol: Yes.  I'm looking at having the returned objects from derivatives have a mime_type method.

README.md
escowles: I would greatly prefer the derivative definitions be in the app's models directory or in an initializer, instead of monkey-patching.  I'm not opposed to monkey-patching when the module is out of your control or can't/won't be changed for some reason.  But that's not the case here and we should be able to provide a cleaner extension mechanism.

spec/hydra/works/services/generic_file/persist_derivative_spec.rb
tampakis: I replaced this spec with the more specific ones in https://github.com/projecthydra-labs/hydra-works/blob/derivatives_spec_refactor/spec/hydra/works/services/persist_derivatives_spec.rb

README.md
mjgiarlo: @terrellt raises a good point about inheritance, and I imagine that folks will want to tweak this on a model-by-model basis rather than say "I want markdown derivs for everything."  I prefer the pattern he suggests to the new one I suggested, and it has the advantages my pattern does along with symmetry with all the other code in Hydraland that deals with derivatives (e.g., https://github.com/psu-stewardship/archivesphere/blob/develop/app/models/concerns/archival_derivatives.rb).

lib/hydra/works/models/concerns/generic_file/contained_files.rb
mjgiarlo: I believe Sufia only includes these 4 if you've explicitly turned on audiovisual processing (which is off by default). I'm in favor of striking these four lines and creating a ticket in Sufia::Core that says "Add these back here at this level because they're not a core Hydra::Works concern" or some such.  Reasonable?

lib/hydra/works/models/concerns/generic_file/derivatives.rb
mjgiarlo: :+1:

spec/hydra/works/services/generic_file/persist_derivative_spec.rb
mjgiarlo: :ok_hand:

lib/hydra/works/models/concerns/generic_file/contained_files.rb
jcoyne: I don't think this will work because all of these have the same type. Doesn't directly_contains_one depend on a unique type to differentiate which is which?

lib/hydra/works/services/generic_file/add_file_to_generic_file.rb
```diff
@@ -48,21 +48,23 @@ def persist
...
         # @param object for mimetype to be determined. Attempts to use methods: :mime_type, :content_type, and :path.
-        def determine_mime_type(file)
+        def determine_mime_type(file, mime_type)
           if file.respond_to? :mime_type
             return file.mime_type
           elsif file.respond_to? :content_type
             return file.content_type
           elsif file.respond_to? :path
             return Hydra::PCDM::GetMimeTypeForFile.call(file.path)
+          elsif mime_type
```
jcoyne: If the mime_type is provided shouldn't it have precedence over the result from `GetMimeTypeForFile`, that's basically a wild ass guess based on file extension.

tampakis: We are working on getting rid of providing the mime_type option, but I agree.

lib/hydra/works/models/concerns/generic_file/contained_files.rb
tampakis: AFAIK the type attribute doesn't need to be unique. These extra service file derivatives are going to remain in Sufia.

jcoyne: @tampakis otherwise there is no way to distinguish which is which, see https://github.com/projecthydra/active_fedora/blob/master/lib/active_fedora/associations/directly_contains_one_association.rb#L20

lib/hydra/works/services/generic_file/persist_derivative.rb
tampakis: Yes this behavior was expected in Sufia according to the [mp3 spec](https://github.com/projecthydra-labs/hydra-works/pull/183/files#diff-8b0e62a1ec782e7f1facccb31b5641caR129) and [ogg spec](https://github.com/projecthydra-labs/hydra-works/pull/183/files#diff-8b0e62a1ec782e7f1facccb31b5641caR143). If the original audio is already in the format of one of its derivatives, this would copy over the original file's bit stream to attach to the object. I'm unsure of what the benefits actually are of doing this. Especially considering the way I implemented that behavior, an mp3/ogg derivative would be created (not saving any time), and we have a duplicate file stream (not saving any space).

Since the audio derivatives are going away from Works, I'll open an issue/question in sufia-core so we can address this.

lib/hydra/works/services/generic_file/persist_derivative.rb
```diff
@@ -9,13 +9,13 @@ class PersistDerivative < Hydra::Derivatives::PersistOutputFileService
...
-    # @param [#read] file the dervivative filestream optionally responds to :mime_type
-    # @option opts [RDF::URI or String] type URI for the RDF.type that identifies the file's role within the generic_file
+    # @param [#read or String] file the derivative filestream optionally responds to :mime_type
+    # @param [String] extract file type symbol (e.g. :thumbnail) from Hydra::Derivatives created destination_name
     # @option opts [Boolean] update_existing when set to false, always adds a new file.  When set to true, performs a create_or_update
     # @option opts [Boolean] versioning whether to create new version entries
 
-    def self.call(object, file, opts={})
-      type = opts.fetch(:type, ::RDF::URI("http://pcdm.org/use#ServiceFile"))
+    def self.call(object, file, destination_name, opts={})
```
grosscol: I don't believe opts can be filled in from the output_file_service.call from Hydra::Derivatives.  Only a custom processor would be able to fill this in now.  As it stood, the only opt that was ever passed was `:mime_type`, and that has been addressed.

Basically, `update_existing` and `versioning` always get set to `true` for persisting derivatives.  I would like to have `versioning` set to false instead.  I would not like to have versions of thumbnails.

terrellt:
This looks okay to me, but I haven't dug too deep into the derivatives stuff. @mjgiarlo @tampakis?

terrellt:
https://coveralls.io/builds/3251106/source?filename=lib%2Fhydra%2Fworks%2Fprocessor.rb#L5 Any idea why this is no longer covered?

mjgiarlo:
:+1:

tampakis:
The processor is no longer needed because of the [RetrieveSourceFile service](https://github.com/projecthydra/hydra-derivatives/blob/master/lib/hydra/derivatives/services/retrieve_source_file_service.rb) in Derivatives. I just updated the gemfiles and gemspecs. Feel free to rebase. I'll be on a plane during the call today.

PR #188, opened 2015-08-06
Changing RDF Vocabulary namespace to projecthydra.org

escowles:
We really shouldn't use "hydra.org", since that's something completely different.


no-reply:
:+1: Should we raise namespace availability as an issue for steering? Seems like there should be a clear policy, if not a plan for vocab hosting.

escowles:
I'll post something to hydra-tech.  I think it would be great to agree on a namespace domain to use consistently, and work toward being able to host the ontologies using the W3C recipes.  IIRC, projecthydra.org is hosted at wordpress.com, so maybe not the best option.

PR #194, opened 2015-08-25
Rubocop autocorrect should transform rspec describe messages with "NOT" and "only" properly

mjgiarlo:


grosscol:
We'll revert this if the rubocop upstream fixes the issue?

mjgiarlo:
We could, but I doubt rubocop-rspec will fix this upstream. They see this as a feature, not a bug, so adding this to our config is how they want folks to deal with this. See their specs for evidence of this:

https://github.com/nevir/rubocop-rspec/blob/master/spec/rubocop/cop/rspec/example_wording_spec.rb#L7-L8

PR #195, opened 2015-08-27
move full text extraction from curation concerns to works.

blancoj:
Move full text extraction from curation concerns to works.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -29,6 +29,7 @@ Metrics/MethodLength:
 Metrics/AbcSize:
   Exclude:
     - lib/hydra/works/services/generic_file/add_file_to_generic_file.rb
+    - lib/hydra/works/services/generic_file/full_text_extraction_service.rb	
     - lib/hydra/works/models/concerns/generic_file/mime_types.rb
     - lib/hydra/works/models/concerns/generic_file/virus_check.rb
 
--- a/lib/hydra/works/errors/full_text_extraction_error.rb
+++ b/lib/hydra/works/errors/full_text_extraction_error.rb
@@ -1,0 +1,5 @@
+module Hydra::Works #:nodoc:
+  # Raised when full text extraction fails.
+  class FullTextExtractionError < StandardError
+  end
+end
--- a/lib/hydra/works/services/generic_file/full_text_extraction_service.rb
+++ b/lib/hydra/works/services/generic_file/full_text_extraction_service.rb
@@ -1,0 +1,38 @@
+module Hydra::Works
+  # Extract the full text from the content using Solr's extract handler
+  class FullTextExtractionService
+    def self.run(generic_file)
+      new(generic_file).extract
+    end
+
+    delegate :original_file, :id, to: :@generic_file
+
+    def initialize(generic_file)
+      @generic_file = generic_file
+    end
+
+    ##
+    # Extract full text from the content using Solr's extract handler.
+    # This will extract text from the file uploaded to generic_file.
+    # The file uploaded to @generic_file can be accessed via :original_file.
+    #
+    # @return [String] The extracted text
+    def extract
+      uri = URI("#{connection_url}/update/extract?extractOnly=true&wt=json&extractFormat=text")
+      req = Net::HTTP.new(uri.host, uri.port)
+      resp = req.post(uri.to_s, original_file.content, 'Content-type' => "#{original_file.mime_type};charset=utf-8", 'Content-Length' => original_file.content.size.to_s)
+      fail "URL '#{uri}' returned code #{resp.code}" unless resp.code == '200'
+      original_file.content.rewind if original_file.content.respond_to?(:rewind)
+      JSON.parse(resp.body)[''].rstrip
+
+    rescue => e
+      raise Hydra::Works::FullTextExtractionError.new, "Error extracting content from #{id}: #{e.inspect}"
+    end
+
+    private
+
+    def connection_url
+      ActiveFedora.solr_config[:url]
+    end
+  end
+end
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -32,6 +32,11 @@ module Hydra
       autoload :GenerateThumbnail
       autoload :UploadFileToGenericFile
       autoload :PersistDerivative
+      autoload :FullTextExtractionService
+    end
+
+    autoload_under 'errors' do
+      autoload :FullTextExtractionError
     end
   end
 end
--- a/spec/hydra/works/services/full_text_extraction_service_spec.rb
+++ b/spec/hydra/works/services/full_text_extraction_service_spec.rb
@@ -1,0 +1,26 @@
+require 'spec_helper'
+
+describe Hydra::Works::FullTextExtractionService do
+  let(:generic_file) { Hydra::Works::GenericFile::Base.new }
+
+  describe '#run' do
+    let(:service_instance) { double }
+    it 'creates an instance of the service and calls .extract on it' do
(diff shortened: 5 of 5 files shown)


.rubocop.yml
```diff
@@ -23,6 +23,7 @@ Metrics/MethodLength:
     - lib/hydra/works/models/concerns/generic_file/virus_check.rb
 
 Metrics/AbcSize:
+  Enabled: false
```
mjgiarlo: Can you remove this change?

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
```diff
@@ -0,0 +1,33 @@
+module Hydra::Works
+  # Extract the full text from the content using Solr's extract handler
+  class FullTextExtractionService
+    def self.run(generic_file)
+      new(generic_file).extract
+    end
+
+    delegate :original_file, to: :@generic_file
+
+    def initialize(generic_file)
```
mjgiarlo: Can you make the methods below this line private to clarify what's in and not in the public API of this service?

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
```diff
@@ -0,0 +1,33 @@
...
+      @generic_file = generic_file
+    end
+
+    def extract
+      uri = URI("#{connection_url}/update/extract?extractOnly=true&wt=json&extractFormat=text")
+      req = Net::HTTP.new(uri.host, uri.port)
+      resp = req.post(uri.to_s, original_file.content,           'Content-type' => "#{original_file.mime_type};charset=utf-8",
```
mjgiarlo: The spacing and indenting looks funky here.

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
```diff
@@ -0,0 +1,33 @@
...
+                                                                 'Content-Length' => original_file.content.size.to_s)
+      fail "URL '#{uri}' returned code #{resp.code}" unless resp.code == '200'
+      original_file.content.rewind if original_file.content.respond_to?(:rewind)
+      JSON.parse(resp.body)[''].rstrip
+
+      # logger is not available in works, so this won't be done.
```
mjgiarlo: It's OK if logging isn't made available in the hydra-works service.  Maybe we should keep the `rescue => e` in here, and then re-raise `e` as a `Hydra::Works::FullTextExtractionError` (which you'd need to create) that can be rescued by consumers to feed a logger. If so, you can delete all the commented out code here.

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
```diff
@@ -0,0 +1,33 @@
+module Hydra::Works
+  # Extract the full text from the content using Solr's extract handler
+  class FullTextExtractionService
+    def self.run(generic_file)
```
mjgiarlo: Can you add lightweight YARD documentation to this public method since it's the main entry point to the service?

mjgiarlo:
Thanks much for the PR, @blancoj!  I have just a few comments and requests for you.

.rubocop.yml
blancoj: We put this in because if it's not in, rubocop fails.  It exist presently in curation concersn:

https://github.com/projecthydra-labs/curation_concerns/blob/50966373f857ddc9a04e8ef79904b00ccd73650b/.rubocop.yml#L20

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
blancoj: private below this line cause spec to fail.  I did make connection_url private.

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
blancoj: The spacing was done by rubocop.  I also thought it was funky.

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
blancoj: I gave it a go.  Let me know what you think.

.rubocop.yml
mjgiarlo: Instead of disabling it altogether, can you add file-specific excludes below?  When you remove this line and run rubocop, what does the output look like?  Feel free to gist that and ping me.

grosscol:
In order to support the full text extraction functionality, solr needs additional libraries.  Obtaining those libraries was done by an install generator in curation concerns.  The issue of installing the libraries for full text extraction, and modifying the solr config were eschewed here, as those changes have already been made in umich's shared development environment.

So the issue of getting the libraries and modifying the solrconfig.xml remains.  I created an issue to gauge interest in putting as an option in jettywrapper with projecthydra/jettywrapper#46.

A number of libraries need to be added under lib/contrib/extraction/lib, and the needs to be modified to point to them solrconfig.xml
```xml
<!-- for full-text indexing -->
<lib dir="../lib/contrib/extraction/lib" regex=".*\.jar" />
```

lib/hydra/works/errors/full_text_extraction_error.rb
```diff
@@ -0,0 +1,5 @@
+module Hydra::Works #:nodoc:
+  # Generic ActiveFedora exception class
```
mjgiarlo: Comment doesn't seem to match up with the class, so you can kill the comment.

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
mjgiarlo: This is a good start. Could you reformat it as YARD? See here for an example: https://github.com/projecthydra/hydra-collections/pull/84/files#diff-68cd07172cdadf2489d69b510b06b313R23

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
mjgiarlo: Can you remove the extraneous extra spaces and put the whole method invocation on one line? That should fix it up. (It'll be a long line, but that's not an egregious sin and the spacing will no longer trip folks up.)

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
```diff
@@ -0,0 +1,40 @@
...
+      uri = URI("#{connection_url}/update/extract?extractOnly=true&wt=json&extractFormat=text")
+      req = Net::HTTP.new(uri.host, uri.port)
+      resp = req.post(uri.to_s, original_file.content,           'Content-type' => "#{original_file.mime_type};charset=utf-8",
+                                                                 'Content-Length' => original_file.content.size.to_s)
+      fail "URL '#{uri}' returned code #{resp.code}" unless resp.code == '200'
+      original_file.content.rewind if original_file.content.respond_to?(:rewind)
+      JSON.parse(resp.body)[''].rstrip
+
+    rescue => e
+      raise Hydra::Works::FullTextExtractionError.new, "Error extracting content from #{id}: #{e.inspect}"
```
mjgiarlo: :+1:

mjgiarlo:
Left some comments that shouldn't add too much work, @blancoj. Thanks!

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
atz: Sometimes you have run rubocop multiple times as it will detect subsequent issues causes by its fixes to earlier ones.

Issue #89, opened 2015-05-13
Move file Characterization from Sufia to Hydra::Works


flyingzumwalt:
Move Sufia::GenericFile::Characterization to Hydra::Works::File::Characterization

flyingzumwalt:
@jcoyne during the sprint it sounded like this was going to necessarily come along with #90 (Derivative Generation) but I don't see it in the code base anywhere.  Was it ruled out?  If so, I need to move it into curation_concerns.

jcoyne:
@flyingzumwalt No idea. Is there some reason you're asking me?

awead:
@flyingzumwalt I decided to punt on characterization because it was a bit much at the time and required getting FITS in the mix. Also, we haven't mapped out FITS to RDF yet, which I think should happen when we add the characterization service to hydra-works.

mjgiarlo:
@awead And that work is coming along, but not yet done: sufia#1006

escowles:
@mjgiarlo Any progress on https://github.com/projecthydra/sufia/issues/1006 since the flurry of discussion about the RDF vocab in early June?  As long as we're moving characterization down into hydra-works, it seems like a good time to RDF-ify it too.

mjgiarlo:
@escowles In light of a potential discussion about related files at next week's Hydra Tech call, and some reticence from folks about the utility of RDFifing FITS output, I was hoping to abandon the effort, tbh. I'm planning to re-assess after we discuss related files. If you have a particular interest in RDF-ifying FITS output, though, I don't think there's any real downside to it, and I'd be happy to hand it off. Your call?

mjgiarlo:
Planning to include this in a version > 0.1 of Hydra-works.

mjgiarlo:
@grosscol is working on this, I believe.

jcoyne:
Currently work on this is in https://github.com/projecthydra-labs/hydra-works/issues/209

mjgiarlo:
@jcoyne @grosscol Can this be closed? If so, lines 48 and 165 could use some attention here: https://github.com/projecthydra-labs/curation_concerns/blob/master/spec/controllers/curation_concerns/file_sets_controller_spec.rb

grosscol:
Made a new issue for the pending specs: projecthydra-labs/curation_concerns#373

PR #199, opened 2015-08-28
Characterization sans om

grosscol:
Still a work in progress.

### DIFF
--- a/lib/hydra/works/models/concerns/generic_file/characterization.rb
+++ b/lib/hydra/works/models/concerns/generic_file/characterization.rb
@@ -1,0 +1,95 @@
+include RDF
+
+module Hydra::Works::GenericFile
+  module Characterization
+    extend ActiveSupport::Concern
+
+    autoload :SaxDoc, 'hydra/works/models/generic_file/characterization/sax_doc'
+
+    DEFAULT_FITS_TO_RDF = { filename: "filename",
+                            mime_type: ["identity", "toolname=FITS", "mimetype"],
+                            file_size: "size",
+                            well_formed: "well-formed",
+                            valid: "valid",
+                            last_modified: ["lastmodified", "toolname=Exiftool"],
+                            date_created: ["created", "toolname=Exiftool"],
+                            fits_version: %w(fits version),
+                            exif_version: ["tool", "toolname=Exiftool", "toolversion"],
+                            original_checksum: "md5checksum" }
+
+    # Setter for instance parse_config
+    def parse_config=(cfg)
+      @parse_config = cfg
+    end
+
+    # Instance method for obtaining the parse_config where instance trumps class if present.
+    def parse_config
+      @parse_config || self.class.parse_config
+    end
+
+    included do
+      include Hydra::Derivatives::ExtractMetadata
+
+      # Base characterization properties
+
+      property :filename, predicate: RDF::Vocab::EBUCore.filename, multiple: false
+      property :file_size, predicate: RDF::Vocab::EBUCore.fileSize, multiple: false
+      property :well_formed, predicate: RDF::URI.new("http://projecthydra.org/ns/fits/wellFormed"), multiple: false
+      property :valid, predicate: RDF::URI.new("http://projecthydra.org/ns/fits/valid"), multiple: false
+      property :date_created, predicate: RDF::Vocab::EBUCore.dateCreated, multiple: false
+      property :last_modified, predicate: RDF::Vocab::EBUCore.dateModified, multiple: false
+      property :fits_version, predicate: RDF::Vocab::PREMIS.hasCreatingApplicationVersion
+      property :exif_version, predicate: RDF::Vocab::EXIF.exifVersion
+      property :original_checksum, predicate: RDF::Vocab::PREMIS.hasMessageDigest, multiple: false
+      # ebucore:hasMimeType multiple: false
+      # property :mime_type, delegate_to: 'original_file'
+      property :mime_type, predicate: RDF::Vocab::EBUCore.hasMimeType, multiple: false
+
+      # Use rails class_attribute which is inherited by subclasses, and independently mutable by them.
+      class_attribute :parse_config
+      self.parse_config = ::Hydra::Works::GenericFile::Characterization::DEFAULT_FITS_TO_RDF.dup
+
+      # Extract the metadata from a content file and store it as properties.
+      # The target properties must already exist.
+      # @param [ActiveFedora::File] file to characterize
+      def characterize(file = original_file)
+        # Run fits and get xml string
+        extracted_md = extract_metadata(file)
+        # Parse xml string to get properties and values
+        prop_values = parse_metadata(extracted_md)
+        # Store metadata as properties of the file.
+        # store_metadata(file.metadata_node, extract_metadata)
+        store_metadata(prop_values)
+      end
+
+      protected
+
+      def extract_metadata(file = original_file)
+        return unless file.has_content?
+        Hydra::FileCharacterization.characterize(file.content, file.original_name, :fits) do |cfg|
+          cfg[:fits] = Hydra::Derivatives.fits_path
+        end
+      end
+
+      def parse_metadata(md)
+        parser = Nokogiri::XML::SAX::Parser.new(Hydra::Works::GenericFile::Characterization::SaxDoc.new(parse_config))
+        parser.parse(md)
+        parser.document.property_values
+      end
+
+      # Assign values of the instance properties from the metadata mapping :prop => val
+      # Multi-valued properties are arrays. Single valued property cannot be appended to.
+      def store_metadata(metadata)
+        metadata.each_pair do |prop, val|
+          next unless self.respond_to? prop
+          property = send(prop)
+          if property.is_a? Array
+            property << val
+          else
+            val.is_a?(Array) ? send("#{prop}=", val.first) : send("#{prop}=", val)
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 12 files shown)


mjgiarlo:
Can you squash your commits, @grosscol?

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+
+    subject { demoClass.new(id:'demo/01/01') }
+
+    before do
+      # Mocking out the Hydra::Works::GenericFile sufficiently to add content, yet not require
+      # a save to fedora
+      allow(file).to receive(:original_name).and_return(filename)
+      allow(file).to receive(:mime_type).and_return(mimetype)
+      # Mock ldp_source.head so AF association believes subject is already saved.
+      allow(subject.ldp_source).to receive(:head).and_return("thefavor")
```
grosscol: This allows for adding the file to generic file without calling save.  It circumvents this by causing the check on if an object is new to return false.  This struck me as a bad plan to be monkeying about with, but I wasn't sure about the extent of the badness.

I might go so far as to mock out a more legitimate `:head` attribute for the `lpd_source` of the GenericFile based on what ActiveFedora would actually stuff in there.

jcoyne:
I need height and width for images and videos. Does this discard that?

grosscol:
@jcoyne should height and width be included in the default config of metadata to extract?

Currently this just works by a simple mapping which is barely more complex than a hash of property values => element names.  The config works well enough for parsing metadata from fits xml output.

So long as the class has a height and width property and the fits xml contains the data of interest, it should be straightforward to amend a class or instance's config so that it's picked up with a call to characterize.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,235 @@
...
+            end
+          end
+        end
+      end
+
+    end # of inclduded do
+
+    # SAX parsing class for FITS
+    # Requires fits_to_rdf_mapping
+    class SaxDoc < Nokogiri::XML::SAX::Document
```
grosscol: This inner class needs to be extracted to some other file, but I am not sure what a good destination for it would be.

It's a SAX parser that the characterization concern depends on, so `lib/hydra/works/models/concerns/generic_file/characterization/sax_doc.rb` seems like it should be the place.  That would put a class under the concerns directory tree which seems like a bad plan

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
```diff
@@ -0,0 +1,157 @@
...
+
+    let(:filename)     { 'sample-file.pdf' }
+    let(:mimetype)     { 'application/pdf' }
+    let(:file_content) { File.open(File.join(fixture_path, filename)) }
+    let(:file)         { Hydra::PCDM::File.new.tap{ |f| f.content = file_content } }
+    let(:type)         { ::RDF::URI("http://pcdm.org/use#OriginalFile") }
+
+    subject { demoClass.new(id:'demo/01/01') }
+
+    before do
```
grosscol: @terrellt I was mocking out the add file to generic file service so that generic files did not have to be saved in order to have a file attached as content.  This is what I came up with.  My next plan is to try to extract and encapsulate the logic in a spec_helper function, but have not had luck doing that so far.

lib/hydra/works/models/concerns/generic_file/characterization.rb
escowles: Yep, adding a class to the concerns tree does seem bad. Maybe we need a new directory in lib/hydra/works/models for it?

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
escowles: Mocking the interactions like this is a definite improvement.  But I thought the goal was to allow actually working with the unsaved objects, not just get the tests to work without saving.

grosscol: I believe that is the ultimate goal.  To be able to build the relationships in memory, work with them, and possible persist the whole lot back to fedora.  I think that is a much thornier situation.  I am currently wondering about the badness level of mocking the ldp_subject.head response.

escowles: I agree it seems potentially very bad.  But in cases like this where you're not testing the association, it seems reasonable to me.

terrellt: > I am currently wondering about the badness level of mocking the ldp_subject.head response.

It's pretty bad, but only because #head will never -actually- return "thefavor". When you mock something it should return something with a similar API, if not a real possible return value.

tampakis:
Fits doesn't installed in travis. Skip the failing test in Travis?

jcoyne:
Please do a brief write-up on why you're doing this work. Is this an improvement over what we already have?

grosscol:
This PR was in response to #89.  As @jcoyne points out, the methodology of this PR warrants some explanation. Specifically, what is being done here and why.  

The two most glaring changes are:
1. Using simple SAX parser instead of OM to parse fits output to datastream.
2. Using smaller set of properties as the base set instead of more comprehensive list.

##### SAX instead of OM
###### TL;DR It seemed simpler to write a SAX parser for fits.xml than to figure out how to use OM to set the properties directly on the GenericFile.
OM is a very robust gem that can parse any XML, and has extensive configuration that would permit extraction of any element or attribute value.  It uses a DOM parser, is schema aware, and is namespace aware. In short, OM is an all terrain industrial loader.

It's a piece of heavy construction equipment that appears to have been primarily used to drive down to the same fits.xml store.  If the most that is going to be done with characterization is parsing a specific version of fits.sh output, it seems like a lighter solution might be more applicable.  The non-OM solution was a SAX parser with a simple configuration.

Is it better? nope.  I was under the impression that we wanted to get away from fits datastreams in favor of defining properties directly on the object, and this seemed like a reasonable approach for that.

##### Smaller default properties set
###### TL;DR having extra properties that aren't applicable to the content is confusing.
@escowles pointed me to a document produced by the hydra technical metadata group.  It detailed some core technical metadata.  It is a small default set that could be added to as needed by classes further downstream.

Having height and width methods on a GenericFile that contains an audio file was unsettling.  Interrogating an object for the height and width I was not sure if characterization could not determine those, or if they were not supposed to be there in the first place.  So the implementation in this PR punts that decision to the downstream projects and sticks to a minimal set of technical metadata.

escowles:
:+1: to using SAX and having a smaller set of default properties.  I think having a small default set and an example of how to extend it is the right approach, and a good approach for implementors to emulate.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,95 @@
...
+    autoload :SaxDoc, 'hydra/works/models/generic_file/characterization/sax_doc'
+
+    DEFAULT_FITS_TO_RDF = { filename: "filename",
+                            mime_type: ["identity", "toolname=FITS", "mimetype"],
+                            file_size: "size",
+                            well_formed: "well-formed",
+                            valid: "valid",
+                            last_modified: ["lastmodified", "toolname=Exiftool"],
+                            date_created: ["created", "toolname=Exiftool"],
+                            fits_version: %w(fits version),
```
terrellt: Why does this one get a %w and the others are just arrays? (I like arrays, not sure what the standards say).

jcoyne:
@grosscol Now we are adding another piece of code and not really removing anything (OM is a dependency of AF).  And the SAX parser isn't a big benefit on a tiny xml file such as that omitted by Fits.  Most importantly, we are already using several fields defined by the om definition that are not property being captured by this new implementation (image height and width, video height, width and duration, audio duration).  Finally, I'm not seeing any direction on how to extend this approach provided with this PR.

lib/hydra/works/models/concerns/generic_file/characterization.rb
mjgiarlo: There is a Rubocop style that says "only use %w for arrays" but I disabled it straight away. :+1: to arrays with square brackets, especially in this case for consistency's sake.

grosscol:
@jcoyne I was removing the cognitive overhead of sorting out OM.  The terminology configuration was difficult to understand, and is designed around using and keeping xml datastreams.

I agree that DOM vs FITS for small files is half of one six of the other.  I was also working on an approach using OM.  It ends up building an entire OM XML Document only to discard it.  That seemed like building an entire scaffolding when we would only need a ladder.

Making a sub class of Generic file with additional properties and mappings is how I would expect downstream applications to add the properties they require.  I don't see a benefit for having video height on a pdf or audio file.  At best, that's not harmful.

```
IMG_FITS_TO_RDF = { image_width:  ["imageWidth",  "toolname=Jhove"],
                    image_height: ["imageHeight", "toolname=Jhove"],
                    color_space:  ["colorSpace",  "toolname=Jhove"] }

class Demo < Hydra::Works::GenericFile::Base
  include Hydra::Works::GenericFile::Characterization
  property :image_width, predicate: ::RDF::URI.new('http://NeedAPredicate.org/ns/hydra/imageW'), multiple: false
  property :image_height, predicate: ::RDF::URI.new('http://NeedAPredicate.org/ns/hydra/imageH'), multiple: false
  property :jhove_time, predicate: ::RDF::URI.new('http://NeedAPredicate.org/ns/hydra/executionTime'), multiple: false
  parse_config.merge!(IMG_FITS_TO_RDF)
end
```

I have a branch where I was using OM, and then trying to get the values from the resulting OM XML Document.  I can go back to that branch and get it to an equivalent state.

jcoyne:
I'm absolutely not attached to OM. I do think we should support a base list of predicates that can support all types of media because this is commonly for a self deposit system.  We can't say only use image predicates if its an image, because it is FITS itself that tells us so.

PR #200, opened 2015-08-31
Prevents Rubocop from inspecting bin directory and schema.rb in spec/internal

flyingzumwalt:


.rubocop.yml
```diff
@@ -5,6 +5,9 @@ AllCops:
   DisplayCopNames: true
   Include:
     - '**/Rakefile'
+  Exclude:
+    - 'spec/internal/bin/*'
```
grosscol: Why not just ignore spec/internal/* ?

flyingzumwalt: I thought it might be good to catch bad stuff generated into spec/internal by the generators.

mjgiarlo: Dig this idea.

mjgiarlo:
The build is failing because Rubocop is now scanning a bunch of stuff in vendor/bundle/. (I'm not sure why the change you made, @flyingzumwalt, would cause so many new Rubocop failures, though...)

mjgiarlo:
:clap: @cbeer 

Exclude the vendor directory also, @flyingzumwalt, and the build should get farther.

flyingzumwalt:
:arrow_up: Green!  @mjgiarlo or @grosscol could you merge?

PR #204, opened 2015-09-04
add hydra_works:jetty:config task

### LINKED ISSUE #201: Add rake task to handle full-text extraction
The full-text extraction feature requires a bunch of .jar files to handle the actual extraction and also requires a small change to Solr's configuration that makes it aware of the .jar files and turns on Solr's `ExtractingRequestHandler`. This code used to live in a rake task and a generator in Sufia, but now that this feature has moved to hydra-works, and hydra-works does not have generators, it needs to be reimplemented as a hydra-works rake task.

The relevant code and files in Sufia look like this:

* https://github.com/projecthydra/sufia/blob/c47dd7bacc3c3fb1e8b2649165474dd38e4a0e74/sufia-models/lib/tasks/sufia-models_tasks.rake#L37-L108
* https://github.com/projecthydra/sufia/blob/c47dd7bacc3c3fb1e8b2649165474dd38e4a0e74/sufia-models/lib/generators/sufia/models/fulltext_generator.rb#L19-L26
* https://github.com/projecthydra/sufia/blob/c47dd7bacc3c3fb1e8b2649165474dd38e4a0e74/sufia-models/lib/generators/sufia/models/templates/config/solrconfig.xml

This work **also** includes making sure this test runs and passes in the Travis CI environment (currently it is not run on Travis):

https://github.com/projecthydra-labs/hydra-works/blob/master/spec/hydra/works/services/full_text_extraction_service_spec.rb#L15-L25

tampakis:
Closes #201. Rake tasks pulled down from curation_concerns and full_test_extraction service test runs in travis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 /coverage/
 /doc/
 /pkg/
+/solr_conf/
 /spec/reports/
 spec/internal
 /tmp/
--- a/Rakefile
+++ b/Rakefile
@@ -5,6 +5,8 @@ require 'rspec/core/rake_task'
 require 'engine_cart/rake_task'
 require 'rubocop/rake_task'
 
+Dir.glob('lib/tasks/*.rake').each { |r| import r }
+
 desc 'Run style checker'
 RuboCop::RakeTask.new(:rubocop) do |task|
   task.requires << 'rubocop-rspec'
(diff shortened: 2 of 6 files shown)


mjgiarlo:
:+1:

PR #209, opened 2015-09-08
Characterization using OM

grosscol:
A parallel line of development with #199.  Also addresses #89.  

This is a more direct pull down of the OM approach.  @jcoyne, this still requires the downstream application to define the properties they want available for their GenericFile::Base derived class.  However, all the default fits datastream configuration is there by default, so only properties will have to be defined for the GF subclass.

This does not save the fits xml datastream object.  It's a transient artifact of the characterization.

The specs are still a work in progress.  I should be able to address them tomorrow.


mjgiarlo:
Travis is :unamused: and pls squash your commits, sir.

jcoyne:
It's a rubocop style violation

jcoyne:
It's a rubocop style violation.

grosscol:
Sorted out the specs and added some basic properties for some of the major media types.

I don't have the call to characterization triggering an obligate .save call.  I was assuming calling applications might want to do more work or inspect the results before doing a .save.  Is this a reasonable assumption? or should we make the call to characterize implicitly save the generic file ?

jcoyne:
Yes, I think that's a great assumption. Don't save here.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,120 @@
+include RDF
+
+module Hydra::Works::GenericFile
+  module Characterization
+    extend ActiveSupport::Concern
+
+    autoload :FitsDatastream, 'hydra/works/models/generic_file/characterization/fits_datastream.rb'
+    autoload :Image, 'hydra/works/models/concerns/generic_file/characterization/image.rb'
+    autoload :Doc, 'hydra/works/models/concerns/generic_file/characterization/doc.rb'
```
mjgiarlo: Could you rename this module to `Document` for consistency with the other modules? It also makes it consistent with the FITS output. I don't care deeply about this one, but if you wind up having to make other changes...

grosscol: Good catch.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,120 @@
...
+      def extract_metadata(file = original_file)
+        return unless file.has_content?
+        Hydra::FileCharacterization.characterize(file.content, file.original_name, :fits) do |cfg|
+          cfg[:fits] = Hydra::Derivatives.fits_path
+        end
+      end
+
+      # Use OM to parse metadata
+      def parse_metadata(metadata)
+        fdata = parser_class.new
```
mjgiarlo: What does this variable signify? I'd prefer a longer, more descriptive variable name here than a shorter one. (This runs counter to practice in other languages, but is more or less idiomatic in Ruby.)

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,120 @@
...
+
+      # Assign values of the instance properties from the metadata mapping :prop => val
+      def store_metadata(terms)
+        # Go through each term.
+        terms.each_pair do |term, value|
+          # Figure out property to use or nil
+          property = property_for(term)
+          next if property.nil?
+          # Array-ify the value to avoid a conditional here
+          value = [value] unless value.is_a? Array
```
mjgiarlo: Can simplify lines 94 and 96:

```ruby
Array(value).each { |v| append_property_value(property, v) }
```

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,120 @@
...
+
+      property :filename, predicate: RDF::Vocab::EBUCore.filename, multiple: false
+      property :format_label, predicate: RDF::Vocab::PREMIS.hasFormatName
+      property :file_size, predicate: RDF::Vocab::EBUCore.fileSize
+      property :well_formed, predicate: RDF::URI.new("http://projecthydra.org/ns/fits/wellFormed")
+      property :valid, predicate: RDF::URI.new("http://projecthydra.org/ns/fits/valid")
+      property :date_created, predicate: RDF::Vocab::EBUCore.dateCreated
+      property :last_modified, predicate: RDF::Vocab::EBUCore.dateModified
+      property :fits_version, predicate: RDF::Vocab::PREMIS.hasCreatingApplicationVersion
+      property :exif_version, predicate: EXIF.exifVersion
```
mjgiarlo: Why not use `RDF::Vocab` for EXIF?

lib/hydra/works/models/concerns/generic_file/characterization/audio.rb
```diff
@@ -0,0 +1,19 @@
+include RDF
+
+module Hydra::Works::GenericFile::Characterization
+  module Audio
+    extend ActiveSupport::Concern
+    included do
+      property :bit_depth, predicate: Vocab::NFO.bitDepth
```
mjgiarlo: I believe you can use `RDF::Vocab` for all the NFO and EXIF properties here.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,120 @@
...
+
+      def append_property_value(property, value)
+        property_value = send(property)
+        if property_value.is_a? Array
+          property_value << value
+        else
+          send("#{property}=", value)
+        end
+      end
+    end # of inclduded do
```
mjgiarlo: Typo.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,120 @@
...
+        fdata.ng_xml = metadata if metadata.present?
+        characterization_terms(fdata)
+      end
+
+      # Get terms and values
+      # This is using the names of the proxy terms; not the name of what the proxy points to.
+      def characterization_terms(datastream)
+        h = {}
+        datastream.class.terminology.terms.each_pair do |key, target|
+          # a key is a proxy if it's target responds to proxied_term
```
mjgiarlo: `s/it's/its/`

lib/hydra/works/models/concerns/generic_file/characterization.rb
grosscol: Excellent.  I was unaware that call would flatten existing arrays.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,120 @@
...
+      property :format_label, predicate: RDF::Vocab::PREMIS.hasFormatName
+      property :file_size, predicate: RDF::Vocab::EBUCore.fileSize
+      property :well_formed, predicate: RDF::URI.new("http://projecthydra.org/ns/fits/wellFormed")
+      property :valid, predicate: RDF::URI.new("http://projecthydra.org/ns/fits/valid")
+      property :date_created, predicate: RDF::Vocab::EBUCore.dateCreated
+      property :last_modified, predicate: RDF::Vocab::EBUCore.dateModified
+      property :fits_version, predicate: RDF::Vocab::PREMIS.hasCreatingApplicationVersion
+      property :exif_version, predicate: EXIF.exifVersion
+      property :original_checksum, predicate: RDF::Vocab::PREMIS.hasMessageDigest
+      # property :mime_type, delegate_to: 'original_file'
```
mjgiarlo: Delete?

lib/hydra/works/models/concerns/generic_file/characterization/doc.rb
```diff
@@ -0,0 +1,26 @@
...
+      property :line_count, predicate: RDF::Vocab::NFO.lineCount
+      property :character_set, predicate: RDF::URI.new("http://www.w3.org/2011/content#characterEncoding")
+      property :markup_basis, predicate: RDF::URI.new("http://www.w3.org/2011/content#doctypeName")
+      property :markup_language, predicate: RDF::URI.new("http://www.w3.org/2011/content#systemId")
+      # properties without canonical URIs
+      property :paragraph_count, predicate: RDF::URI.new("http://projecthydra.org/ns/odf/paragraphCount")
+      property :table_count, predicate: RDF::URI.new("http://projecthydra.org/ns/odf/tableCount")
+      property :graphics_count, predicate: RDF::URI.new("http://projecthydra.org/ns/odf/graphicsCount")
+
+      parser_config.merge!(file_title: :title,
```
mjgiarlo: What's this do?

lib/hydra/works/models/concerns/generic_file/characterization/image.rb
```diff
@@ -0,0 +1,26 @@
+include RDF
+
+module Hydra::Works::GenericFile::Characterization
+  module Image
+    extend ActiveSupport::Concern
+    included do
+      property :byte_order, predicate: RDF::URI('http://sweet.jpl.nasa.gov/2.2/reprDataFormat.owl#byteOrder'), multiple: false
+      property :compression, predicate: RDF::URI('http://projecthydra.org/ns/mix/compressionScheme'), multiple: false
+      property :width, predicate: EXIF.imageWidth, multiple: false
```
mjgiarlo: Use `RDF::Vocab` for EXIF properties.

lib/hydra/works/models/concerns/generic_file/characterization/video.rb
```diff
@@ -0,0 +1,20 @@
+include RDF
+
+module Hydra::Works::GenericFile::Characterization
+  module Video
+    extend ActiveSupport::Concern
+    included do
+      property :height, predicate: EXIF.imageLength
```
mjgiarlo: Use `RDF::Vocab` for EXIF.

lib/hydra/works/models/concerns/generic_file/characterization/video.rb
```diff
@@ -0,0 +1,20 @@
+include RDF
```
mjgiarlo: This is a pattern that I don't often see in Hydra code. I have a preference for striking all the "include RDF" lines in the characterization modules and then using the explicit `RDF::Vocab::NFO` (for instance) below and in the other files where this pattern is used.

spec/fixtures/fits_0.8.5_jp2.xml
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mjgiarlo: Does this PR handle both FITS 0.6 and 0.8 equally well? Because that'd be super cool.

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
```diff
@@ -0,0 +1,182 @@
+require 'spec_helper'
+
+# Mocking out the Hydra::Works::GenericFile sufficiently to add original_file
```
mjgiarlo: I have a slight preference to see this code (lines 3-22) put in a helper in `spec/support/` and `require`d at the top of this file.

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+
+describe Hydra::Works::GenericFile::Characterization do
+  let(:demo_class) do
+    Class.new(Hydra::Works::GenericFile::Base) do
+      include Hydra::Works::GenericFile::Characterization
+    end
+  end
+
+  # The is the only test context that will not mock out Characterization.characterize
+  #   Remove after development as this is more of an integration test
```
mjgiarlo: Mark with "TODO" in the comment if it needs further work.

lib/hydra/works/models/concerns/generic_file/characterization.rb
grosscol: Another good catch.  That was leftover terminology from a different implementation.  It should be a datastream, and I'll rename it as such.

The thinking here is that it should be easier to support fits versions by writing a OmDatastream to match.  That way the parser_class could be selected by the downstream app... if they were married to some version of fits.  It's something on my board to do or write a feature issue.

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
```diff
@@ -0,0 +1,182 @@
...
+  context "using fits.sh via Hydra::FileCharacterization" do
+    let(:filename)     { 'sample-file.pdf' }
+    let(:mimetype)     { 'application/pdf' }
+    let(:file_content) { File.open(File.join(fixture_path, filename)) }
+    let(:file)         { Hydra::PCDM::File.new.tap { |f| f.content = file_content } }
```
mjgiarlo: You shouldn't need the `tap` here because `Hydra::PCDM::File` is a subclass of `ActiveFedora::File` which yields `self` to a block passed to the `initialize` method.

lib/hydra/works/models/concerns/generic_file/characterization.rb
grosscol: It was copy pasta from someone's prior work finding predicates for these properties.  I'll update to use RDF::Vocab.

lib/hydra/works/models/concerns/generic_file/characterization.rb
mjgiarlo: I dig it. I thought that we were generally moving away from the 'datastream' terminology since that's not a thing in Fedora 4. Maybe @jcoyne can share an alternative word here.

lib/hydra/works/models/concerns/generic_file/characterization.rb
mjgiarlo: Yup! Pass it an array and it returns the original array. Pass it a non-array and it array-ifies it safely.

(`Array.wrap` is even safer but I believe that's from ActiveSupport.)

lib/hydra/works/models/concerns/generic_file/characterization.rb
mjgiarlo: :clap:

spec/fixtures/fits_0.8.5_jp2.xml
grosscol: It handles 0.6 kinda well.  It was on my list for things to get working perfectly, but that can wait for another time.

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
grosscol: Initially, I did try that.  the `mock_add_file_to_generic_file` function threw errors when I put it in the spec_helper.

mjgiarlo: Here's a (slightly more complicated than you need) example of how to make this work.

First, write your function and an rspec config include in e.g. `spec/support/fedora_mocks.rb`. Example: https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/spec/support/proxies.rb#L2-L13

Then require it in your spec_helper. (This example happens in two steps to combine all feature spec-related helpers together, but you wouldn't need both steps: https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/spec/spec_helper.rb#L23 and https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/spec/support/features.rb#L5 You could just `require 'support/fedora_mocks'` in spec_helper.)

Then call your neatly packaged method: https://github.com/projecthydra/sufia/blob/a064489d3cf4d228abf978d7126644e2c653aa5e/spec/features/proxy_spec.rb#L14

lib/hydra/works/models/concerns/generic_file/characterization.rb
grosscol: It was originally the parsed file data, but the approach of this branch actually does use an OmDatastream to do the parsing.  It just throws it away when it's done instead of keeping it around and delegating properties to it.

lib/hydra/works/models/concerns/generic_file/characterization.rb
grosscol: This is actually a TODO.  The `original_file.mime_type` can currently be different than the `generic_file.mime_type` but I couldn't sort out the syntax for `delegate_to:` that would get that working.

lib/hydra/works/models/concerns/generic_file/characterization/doc.rb
grosscol: Updates the class configuration for mapping parsed terms to the property values.  Added a comment to explain that.

lib/hydra/works/models/concerns/generic_file/characterization.rb
mjgiarlo: :OK: Can you include that in a comment above that commented-out code?

mjgiarlo: thx!

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,116 @@
+module Hydra::Works::GenericFile
+  module Characterization
+    extend ActiveSupport::Concern
+
+    autoload :FitsDatastream, 'hydra/works/models/generic_file/characterization/fits_datastream.rb'
```
jcoyne: If you extend ActiveSupport::Autoload, then you don't need the second arguments on any of these calls.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,116 @@
...
+      property :format_label, predicate: RDF::Vocab::PREMIS.hasFormatName
+      property :file_size, predicate: RDF::Vocab::EBUCore.fileSize
+      property :well_formed, predicate: RDF::URI.new("http://projecthydra.org/ns/fits/wellFormed")
+      property :valid, predicate: RDF::URI.new("http://projecthydra.org/ns/fits/valid")
+      property :date_created, predicate: RDF::Vocab::EBUCore.dateCreated
+      property :last_modified, predicate: RDF::Vocab::EBUCore.dateModified
+      property :fits_version, predicate: RDF::Vocab::PREMIS.hasCreatingApplicationVersion
+      property :exif_version, predicate: RDF::Vocab::EXIF.exifVersion
+      property :original_checksum, predicate: RDF::Vocab::PREMIS.hasMessageDigest
+      # property :mime_type, delegate_to: 'original_file'
```
jcoyne: Please remove this.

lib/hydra/works/models/concerns/generic_file/characterization.rb
```diff
@@ -0,0 +1,116 @@
...
+
+      def append_property_value(property, value)
+        property_value = send(property)
+        if property_value.is_a? Array
+          property_value << value
+        else
+          send("#{property}=", value)
+        end
+      end
+    end # of included do
```
jcoyne: cruft?

lib/hydra/works/models/concerns/generic_file/characterization/audio.rb
```diff
@@ -0,0 +1,17 @@
+module Hydra::Works::GenericFile::Characterization
+  module Audio
+    extend ActiveSupport::Concern
+    included do
+      property :bit_depth, predicate: RDF::Vocab::NFO.bitDepth
+      property :channels, predicate: RDF::Vocab::NFO.channels
+      property :data_format, predicate: RDF::Vocab::EBUCore.hasDataFormat
+      property :frame_rate, predicate: RDF::Vocab::NFO.frameRate
+      property :duration, predicate: RDF::Vocab::NFO.duration
+      property :height, predicate: RDF::Vocab::EXIF.imageLength
```
jcoyne: :clap: Thanks!

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+    end
+  end
+
+  context "using audio module" do
+    let(:characterization)     { class_double("Hydra::FileCharacterization").as_stubbed_const }
+    let(:fits_filename)        { 'fits_0.8.5_mp3.xml' }
+    let(:fits_response)        { IO.read(File.join(fixture_path, fits_filename)) }
+    let(:file)                 { Hydra::PCDM::File.new }
+    let(:video_class) do
+      Class.new(Hydra::Works::GenericFile::Base) do
```
jcoyne: I don't understand why we would do this? We don't know the GenericFile is an audio until *after* we do the characterization.  We need the Audio module mixed in to GenericFile all the time.

grosscol: An application can still have an everything bagel.  If the repo can expect all of the types, as curation concerns does, including all four media type modules will be the course of action.

I went with the a la cart model because the application may have more apriori information about an uploaded file or the nature of the repository.  For example, HathiTrust doesn't take Audio or Video so having those properties would be clutter at best.

Or it may be known that only the simplest characterization will be useful because the files that you're going to be ingesting may have been emitted by some proprietary scientific appliance or software package.  Those files can be pathological in that they'll appear to be a valid file type but interact badly or unexpectedly with other software.

lib/hydra/works/models/concerns/generic_file/characterization.rb
grosscol: I was unable to specify that the rails autoload logic should be looking in the models directory.

lib/hydra/works/models/concerns/generic_file/characterization.rb
grosscol: The comment or the two short functions above?  The comment was to keep my place when the module had a mix of class and instance methods.  The functions were my response to rubocop calling out functions that were too verbose or too complex.

Comment removed.  Not willing to shoot it out with rubocop about the functions again.

spec/fixtures/fits_0.8.5_jp2.xml
mjgiarlo: :clap:

lib/hydra/works/models/concerns/generic_file/characterization.rb
mjgiarlo: Hydra-works doesn't declare Rails as a dependency, I thought.

jcoyne: @mjgiarlo Hydra-Works depends on ActiveFedora which depends on ActiveSupport.  Yes, that models makes it tricky. I think we have a structural problem. A module in `concerns/` probably should not define other modules that are not under `concerns/`.  For the rest of them (Image, Document, Video, Audio), it should be autoloading relative to the current file, so you shouldn't have to worry about it.

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
jcoyne: I guess we're down to philosophy now. My thought is that hydra-works should support a large number of use cases without modification.  This is how it works right now.  If HathiTrust doesn't need everything, then they can define a stripped down version of GenericFile. Want to put this to the hydra-tech list for a discussion?

grosscol: Putting this on the hydra-tech list for discussion sounds like a good place to address the design issue.

.rubocop.yml
```diff
@@ -70,3 +75,8 @@ RSpec/FilePath:
 
 RSpec/InstanceVariable:
   Enabled: false
+
+RSpec/DescribedClass:
+  Exclude:
+    - 'spec/hydra/works/models/concerns/generic_file/characterization_spec.rb'
```
jcoyne: why is this excluded. It looks like it conforms to me.

jcoyne:
Is there some reason this isn't going in Hydra::FileCharacterization?

.rubocop.yml
grosscol: Good catch.  That was from a previous iteration of the characterization spec.

grosscol:
@jcoyne This probably should go in Hydra::FileCharacterization.  I was unaware of that and it was not mentioned in the issue (#89) I was working on.

.rubocop.yml
grosscol: Rubocop will complain about not using described_class in the spec.  However, using described_class doesn't permit this section of the spec to run:
```ruby
describe Hydra::Works::GenericFile::Characterization do
  let(:demo_class) do
    Class.new(Hydra::Works::GenericFile::Base) do
      include Hydra::Works::GenericFile::Characterization
    end
  end
```

spec/hydra/works/models/concerns/generic_file/characterization_spec.rb
```diff
@@ -0,0 +1,286 @@
+require 'spec_helper'
+require 'support/generic_file_helper'
+
+describe Hydra::Works::GenericFile::Characterization do
+  let(:demo_class) do
+    Class.new(Hydra::Works::GenericFile::Base) do
+      include Hydra::Works::GenericFile::Characterization
```
jcoyne: include described_class

grosscol: In the context of let's block, `described_class` is undefined.  
```
Failure/Error: include described_class
     NameError:
       undefined local variable or method `described_class' for #<Class:0x007feea58766a8>
     # ./spec/hydra/works/models/concerns/generic_file/characterization_spec.rb:8:in `block (3 levels) in <top (required)>'
```
It's the reason there is a rubocop exception for RSpec/DescribedClass that excludes this spec.

lib/hydra/works/models/concerns/generic_file/characterization/base.rb
```diff
@@ -0,0 +1,97 @@
...
+                             file_mime_type: :mime_type }
+
+      # Parser class is an OM terminology.
+      # It maps the characterization tool's output xml to terms and values.
+      class_attribute :parser_class
+      self.parser_class = FitsDatastream
+
+      # Extract the metadata from a content file and store it as properties.
+      # @param [ActiveFedora::File] file to characterize
+      def characterize(file = original_file)
```
jcoyne: I'd prefer we leave this as a service, in that way we can characterize the original file without having to retrieve the original file from the repository. Furthermore we don't want to bloat the "GenericFile" api.

grosscol: That sounds good to me.  I'll refactor this to use the same pattern that hydra-derivatives now uses.

jcoyne: :+1:

terrellt:
@jcoyne I'd love to get this closed out, but I'm not sure about the history here. Can you take a look?

jcoyne:
@terrellt I don't know either. I'm not sure why this is needed/wanted.

terrellt:
@jcoyne Characterization is necessary, and this is part of the desire to move it into hydra works from wherever it was before (sufia?), as per the agreed upon code shredding document months ago.

jcoyne:
Ah, I thought we'd just move https://github.com/projecthydra-labs/curation_concerns/blob/master/curation_concerns-models/app/services/curation_concerns/characterization_service.rb into hydra-works then.  This PR goes far beyond that.

grosscol:
Moving the characterization service as is from curation_concerns was the first iteration.  hydra-works doesn't have an application, so no application level config.  In refactoring it to work without the application level Config, I thought it might be a good time to get away from delegating to fits xml datastream and to rdf properties.

grosscol:
Either way, I'd like to close this ticket to get it off my mind.

PR #218, opened 2015-09-28
full text indexing of a local file

### LINKED ISSUE #217: FullTextExtractionService should be able to be called with a local file path
This will eliminate the need to pull the content from Fedora.

jcoyne:
Allow passing a local filename to the full-text indexing service. This
allows us to do the ingest without pulling a copy from Fedora. Fixes #217

### DIFF
--- a/lib/hydra/works/services/generic_file/full_text_extraction_service.rb
+++ b/lib/hydra/works/services/generic_file/full_text_extraction_service.rb
@@ -1,14 +1,21 @@
 module Hydra::Works
   # Extract the full text from the content using Solr's extract handler
   class FullTextExtractionService
-    def self.run(generic_file)
-      new(generic_file).extract
+    # @param [GenericFile] generic_file
+    # @param [String] file_path for a local file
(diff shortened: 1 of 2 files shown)


mjgiarlo:
:cop: :boom:

lib/hydra/works/services/generic_file/full_text_extraction_service.rb
```diff
@@ -30,19 +37,31 @@ def extract
...
+      @content ||= if file_path
+                     File.open(file_path).read
+                   else
+                     original_file.content
+                   end
+    end
+
     # @return [Hash] the request headers to send to the Solr extract service
     def request_headers
-      { Faraday::Request::UrlEncoded::CONTENT_TYPE => "#{original_file.mime_type};charset=utf-8",
```
mjgiarlo: Can `original_file.mime_type` return `nil` now? If so, do things break? The new code is explicitly OK with returning `nil` and I'm wondering if our test coverage of this condition needs to be expanded. If this does not change current behavior, I'm :+1: on this PR.

PR #219, opened 2015-09-29
Use hydra-derivatives 3.0.0.alpha

jcoyne:


mjgiarlo:
Travis :boom: because :cop: 

Otherwise :+1: if Travis is :green_heart:

jcoyne:
Okay, I've fixed the :cop: issues

PR #224, opened 2015-10-05
Change work predicate from GenericWork to Work

jcoyne:
Also changed GenericWorkBehavior -> WorkBehavior

and associations:
  obj.generic_works -> obj.works
  obj.generic_works= -> obj.works=

And renamed GenericFile to FileSet


spec/hydra/works_spec.rb
```diff
@@ -74,18 +50,6 @@
...
-        expect(pcdm_coll).not_to respond_to(:generic_file?)
-      end
-
-      it 'returns that method is not available for a pcdm object' do
-        expect(pcdm_obj).not_to respond_to(:generic_file?)
-      end
-
-      it 'returns that method is not available for a pcdm file' do
-        expect(pcdm_file).not_to respond_to(:generic_file?)
-      end
```
mjgiarlo: You delete a bunch of `not_to respond_to`s in this PR. Can you say why?

mjgiarlo:
:+1:

spec/hydra/works_spec.rb
jcoyne: Because they aren't testing code that is in this gem.  They are asserting the non-existence of a method in a different gem.  This doesn't seem like a useful test.

PR #227, opened 2015-10-07
Remove mapping of file_title and file_author

### LINKED ISSUE #226: Change default mapping for file and author properties from characterization.
The downstream uses have FileSets with author and title properties.  The ch12n properties should be file_author and file_title so as not to conflict with the existing author and title of the FileSet.

The current implementation where the OM term is remapped from file_title to title was short sighted.  It did not assume the case where a FileSet would already have those methods.  As such, the ch12n values will overwrite the FileSet values.

Simply removing the mapping in the Document module will effect this change.
```
parser_mapping.merge!(file_title: :title,
                            file_author: :author)
```

grosscol:
Fixes #226
Remove custom mapping of file_title and file_author.
This will avoid over writing existing title and author values.

### DIFF
--- a/lib/hydra/works/models/characterization/document_schema.rb
+++ b/lib/hydra/works/models/characterization/document_schema.rb
@@ -1,6 +1,6 @@
 class DocumentSchema < ActiveTriples::Schema
-  property :title, predicate: RDF::DC11.title
-  property :author, predicate: RDF::DC11.creator
+  property :file_title, predicate: RDF::DC11.title
+  property :file_author, predicate: RDF::DC11.creator
   property :page_count, predicate: RDF::Vocab::NFO.pageCount
   property :file_language, predicate: RDF::DC11.language
   property :word_count, predicate: RDF::Vocab::NFO.wordCount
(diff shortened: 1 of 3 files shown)


jcoyne:
This is a feature that you are removing. In Sufia you would get to the metadata page and it would have these values entered from the ch18n

grosscol:
I'm actually reverting to the prior behavior of file_author and file_title instead of author and title.  In curation concerns, the FileSet has both a file_author and author method.  The file_author is from characterization and author is from the descriptive metadata.

If the method already exists for title and author, characterization will happily use these when assigning values.  So if the FileSet had a different title than what ch12n determined from the bitstream, it would get overwritten when ch12n runs.

I am expecting the situation where someone uploads a file which is 'kevin_dataset-new(2)-final.xlsx' but specifies the title as 'final dataset' in the upload form.  I would not like 'final dataset' to get overwritten with 'kevin_dataset-new(2)-final.xlsx' or whatever garbage the office application filled in for the title of it's metadata.

mjgiarlo:
Travis is :unamused:

mjgiarlo:
@jcoyne Are you :ok: with what @grosscol described?

mjgiarlo:
:+1: otherwise

jcoyne:
@mjgiarlo if you're okay, sure.

Issue #221, opened 2015-10-05
Refactor persist derivatives spec to not need generic_file.save


grosscol:
The persist derivatives spec is an integration test of hydra derivatives at the hydra works level.  It should be refactored to not require the use of generic_file.save.

PR #231, opened 2015-10-13
Spec improve 221

grosscol:
resolves issue #221


mjgiarlo:
One failed spec because Travis tries to fire up ffmpeg. Skip that test in Travis env?

grosscol:
Ahh... good catch, @mjgiarlo. Thanks.

jcoyne:
Needs rebase now.

PR #233, opened 2015-10-13
You can't shift onto an ActiveTriples property

jcoyne:
If you do it won't persist. See
https://github.com/ActiveTriples/ActiveTriples/issues/19


mjgiarlo:
:+1:

terrellt:
:cop:

PR #241, opened 2015-10-19
New ordering

tpendragon:


lib/hydra/works/models/concerns/collection_behavior.rb
```diff
@@ -75,10 +75,6 @@ def parents
       member_of
     end
 
-    def in_collections
```
mjgiarlo: Why are we dumping this method?

lib/hydra/works/models/concerns/collection_behavior.rb
```diff
@@ -67,7 +67,7 @@ def file_set?
     deprecation_deprecate :works_generic_file?
 
     def member_of
```
mjgiarlo: Can you say more about this change?

lib/hydra/works/models/concerns/work_behavior.rb
```diff
@@ -21,10 +21,10 @@ module WorkBehavior
     included do
       type [Hydra::PCDM::Vocab::PCDMTerms.Object, Vocab::WorksTerms.Work]
 
-      alias_method :generic_works, :works
+      alias_method :generic_works, :ordered_works
```
mjgiarlo: Does this imply that ordered works and filesets are the new default behavior?

spec/hydra/works/models/collection_spec.rb
```diff
@@ -15,46 +15,53 @@
 
   describe '#collections' do
     it 'returns empty array when only works are aggregated' do
-      subject.members << generic_work1
-      subject.members << generic_work2
+      subject.ordered_members << generic_work1
```
mjgiarlo: Is it no longer useful to test `members <<` (in addition to `ordered_members <<`) at this layer?

mjgiarlo:
Oops, just noticed this was already merged.

lib/hydra/works/models/concerns/collection_behavior.rb
jcoyne: Yes, this is a good catch. @tpendragon what's going on here?

jcoyne: I'm guessing because there's already an in_collections method defined by PcdmBehavior https://github.com/projecthydra-labs/hydra-pcdm/blob/master/lib/hydra/pcdm/models/concerns/pcdm_behavior.rb#L55-L57

PR #253, opened 2015-12-07
Remove the tests for related_object. That code is in Hydra::PCDM

jcoyne:


spec/hydra/works/models/collection_spec.rb
```diff
@@ -1,229 +1,87 @@
...
       end
     end
   end
 
-  describe "#ordered_work_ids" do
-    it "returns IDs of ordered works" do
-      subject.ordered_members << generic_work1
-      expect(subject.ordered_work_ids).to eq [generic_work1.id]
+  describe '#ordered_works' do
+    subject { collection.works }
```
mjgiarlo: I may be misreading this, but this block looks identical to the code above on lines 30-35.  Should the subject here be `collection.ordered_works` instead?

mjgiarlo:
:+1: happy to :shipit: when I see :green_heart:

PR #254, opened 2015-12-07
Remove the interstitial nodes before deleting a FileSet

jcoyne:


spec/hydra/works/models/file_set_spec.rb
```diff
@@ -41,4 +41,18 @@
...
+
+  describe '#destroy' do
+    let(:generic_work) { Hydra::Works::GenericWork.create }
+    before do
+      generic_work.ordered_members << file_set
+      generic_work.save
+    end
+
+    it "Removes the proxy, the list_node and the FileSet" do
+      expect {
```
mjgiarlo: Does this effectively test both of the deletes issued in `remove_from_works`?

jcoyne: No, not really. This is WIP. Mostly for Trey's edification

mjgiarlo: OK

PR #255, opened 2015-12-07
Removed FileSet#file_format

jcoyne:
This was a concern specific to CurationConcerns, so it's been moved
there.  See
https://github.com/projecthydra-labs/curation_concerns/pull/538


mjgiarlo:
:+1:

grosscol:
I think this method is still called by CurationConcerns.  We'll need to create a ticket there to update those calls.

PR #276, opened 2016-03-01
FileSets and Works should barf clearly when a Collection is added as a member

### LINKED ISSUE #259: Unclear error message when attempting to add a collection to a work
To reproduce...
```
col = Hydra::Works::Collection.new(id: 'col-1')
wrk = Hydra::Works::Work.new(id: 'w-1')

wrk.members << col
=> ActiveFedora::AssociationTypeMismatch: #<Collection:0x007fc234ef9950> is not a PCDM object.
```

Although a work can have works and file sets as members and both are PCDM objects, it is not the case that all PCDM objects are works and file sets.  It would be more precise for this message to read...

```
wrk.members << col
=> ActiveFedora::AssociationTypeMismatch: #<Collection:0x007fc234ef9950> is not a Hydra::Works work or fileset.
```

mjgiarlo:
Fixes #259

### DIFF
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -24,6 +24,7 @@ module Hydra
 
     autoload :Characterization
     autoload :NotFileSetValidator
+    autoload :NotCollectionValidator
 
     autoload_under 'models' do
       autoload :Collection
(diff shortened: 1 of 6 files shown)


mjgiarlo:
This solution feels a *bit* heavy for what it's doing, which is spitting out a clearer error message that is supplied by Hydra::PCDM: https://github.com/projecthydra-labs/hydra-pcdm/blob/master/lib/hydra/pcdm/validators/pcdm_object_validator.rb#L2-L8 On the other hand, it follows the patterns established in the `NotFileSetValidator` work, so the codebase is internally consistent.

Thoughts, @jcoyne @tpendragon @escowles @elrayle ?

tpendragon:
Not pretty, but it works, and I don't know a better solution. Sooo...:+1:

PR #278, opened 2016-03-08
Ensure ClamAV scanfile will scan file content.

### LINKED ISSUE #277: Virus check is not checking the content of pcdm files
At https://github.com/projecthydra-labs/hydra-works/blob/094f3e58109c2b6470d64e8d352bd2406b67d48c/lib/hydra/works/models/concerns/file_set/virus_check.rb#L44, writing the `file` is writing the PCDM::File object.  It should be writing the content of that PCDM::File.  As a result, the call to scanfile here: https://github.com/projecthydra-labs/hydra-works/blob/094f3e58109c2b6470d64e8d352bd2406b67d48c/lib/hydra/works/models/concerns/file_set/virus_check.rb#L19 is just scanning a string like `#<Hydra::PCDM::File uri="" >`

While it's very efficient to simply scan a single line of text instead of a potentially large file, it's not effective for finding viruses in the content.

grosscol:
Add eicar antivirus testing file as a fixture to support the possibility of writing an integration test, and to have the standard antivirus testing file handy.

Testing the private method calls is not my first choice, but it beat out having an integration test that required actually calling ClamAV.  A more robust solution would be to tease apart and recompose `detect_viruses` in such a way that the behavior is testable without having to call private methods.

Closes #277


spec/hydra/works/models/concerns/file_set/virus_check_spec.rb
```diff
@@ -50,4 +51,15 @@ def scanfile(path)
       expect(subject.detect_viruses).to eq true
     end
   end
+
+  # TODO: Write test that does not depend on directly calling private methods.
+  # Testing private methods like this is ill advised.
```
escowles: I think the way to avoid calling #local_path_for_file directly is to call #detect_viruses with original_file being a stub that responds to #read but not #path.

grosscol:
Going to wait for rubocop 0.38.1

escowles:
:+1:

PR #282, opened 2016-03-23
Deprecate ordered_file_sets

### LINKED ISSUE #279: Ordered file sets should go away.
https://github.com/projecthydra-labs/hydra-works/blob/1ac8be27dc09d03d90b8ec06b5f11d73ff541674/lib/hydra/works/models/concerns/work_behavior.rb#L54

At LDCX, those present agreed, that if you need ordered things (Pages for example), you should have a Page model.  FileSets should be unordered members of Works.

jcoyne:
FileSets should not have the assumption that they are ordered.  If you
need FileSets in order, create a set of interstial Work nodes that are
ordered (e.g. Page, Component). Fixes #279

### DIFF
--- a/lib/hydra/works/models/concerns/work_behavior.rb
+++ b/lib/hydra/works/models/concerns/work_behavior.rb
@@ -14,6 +14,7 @@ module Hydra::Works
   #   10) Hydra::Works::Work can have access metadata
   module WorkBehavior
     extend ActiveSupport::Concern
+    extend Deprecation
 
     included do
       def self.type_validator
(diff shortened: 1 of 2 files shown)


mjgiarlo:
:+1:

barmintor:
I want to lodge, in the way that I often do, a small protest that while a Page surrogate for a page of a book is not a file, sometimes the thing you are representing in your repository is *actually a file*. There's a use case here both for digital archivists and for simply supporting corporate workflow.

azaroth42:
@barmintor Could you expand the use case for ordered files that don't represent anything?  

Files by themselves don't seem to me to have an inherent order, the order is imposed by some construct above them... which is what this change reflects.

barmintor:
@azaroth42  https://github.com/projecthydra-labs/hydra-pcdm/issues/192

PR #284, opened 2016-03-23
Improving test coverage

escowles:


lib/hydra/works/models/concerns/file_set/mime_types.rb
```diff
@@ -23,11 +23,6 @@ def office_document?
       self.class.office_document_mime_types.include? mime_type
     end
 
-    # TODO: remove?
```
jcoyne: Please alter the commit message to mention this method was removed.

jcoyne:
:+1:

PR #291, opened 2016-04-20
Don't require ActiveFedora::Aggregation anymore

awead:
@barmintor


jcoyne:
Shouldn't you add a dependency on AF >= 9.11

jcoyne:
@awead or requiring hydra-pcdm >= 0.6  will accomplish that transitively

awead:
@jcoyne yes, that makes sense. And we're okay with breaking backwards compatibility? Although there isn't any expectation of that, is there.

awead:
@jcoyne updated.

Issue #290, opened 2016-04-20
Hydra::Works::CharacterizationService placing metadata on FileSet instead of File


awead:
This issue was first reported in projecthydra/sufia#1894

Hydra::Works::CharacterizationService is updating the FileSet object with technical metadata extracted from Fits. This data should go in the File object instead.

awead:
In looking into this, it's going deeper. `Hydra::Works::CharacterizationService` takes an `ActiveFedora::Base` object, but it needs to take a `Hydra::PCDM::File`. We'd need to re-tool Hydra::Works to apply the various schemas to that object, similar to how it's being done in Hydra::PCDM:

https://github.com/projecthydra-labs/hydra-pcdm/blob/master/lib/hydra/pcdm/models/file.rb

scossu:
Is the plan to use the default Fedora fcr:metadata resource to store all technical metadata for a bitstream, or a different resource?

mjgiarlo:
@scossu Yes, the plan is to use `fcr:metadata`.

awead:
Fixed in 6e0dd6cc27ea64af816049b936515a4ef62ddc9b

Issue #287, opened 2016-03-30
FitsDatastream does not exist in the Hydra::Works::CharacterizationService name context 


barmintor:
The default value for opts[:parser_class] needs to be Hydra::Works::Characterization::FitsDatastream instead.

barmintor:
This error occurs when Hydra::Works::Characterization has not been included in a model- it sets the parser_class default in a context where FitsDatastream resolves.

awead:
@barmintor this should be fixed now. Can you confirm?

PR #292, opened 2016-04-26
Characterization metadata on Hydra::PCDM::File

awead:
Ok, here's where I am with projecthydra/sufia#1894 / #290 

The big changes here are:
* characterization can be run with a file or on the existing binary content of Hydra::PCDM::File
* if the binary content isn't present, it'll need to be added (by a service or other) so that the metadata will be persisted
* I'm punting on `:multiple` and `index.as` -- make this a concern of FileSet and it can index the values, singly or otherwise, as needed
* allows each File of a FileSet to collect any and all characterization metadata and then let the FileSet sort out which from which File (if there are more than one) it wants to index

ping @jcoyne @mjgiarlo


awead:
pinging @barmintor @escowles as well, since this will affect curation concerns.

lib/hydra/works/services/characterization_service.rb
```diff
@@ -34,7 +34,7 @@ def characterize
       # Get value from opts hash, object, or use default
       def extract_options(opts)
         parser_mapping = fetch_or_respond(opts, :parser_mapping) || {}
-        parser_class = fetch_or_respond(opts, :parser_class) || FitsDatastream
+        parser_class = fetch_or_respond(opts, :parser_class) || Hydra::Works::Characterization::FitsDatastream
```
grosscol: I think this closes #287

lib/hydra/works/services/characterization_service.rb
```diff
@@ -44,17 +44,11 @@ def fetch_or_respond(opts, key)
...
       def source_to_content
-        if source.is_a? String
-          File.open(source)
-        elsif source.is_a? Symbol
-          s = object.send(source)
-          s.respond_to?(:content) ? s.content : s
-        else
-          source
-        end
+        return object.content if source.nil?
```
grosscol: We'll only be calling characterization on pcdm:files now, correct?  So the technical metadata is going on the pcdm:file that currently has the content, or will end up with the content?

lib/hydra/works/models/characterization/already_there_strategy.rb
```diff
@@ -4,9 +4,9 @@ class AlreadyThereStrategy < ActiveTriples::ExtensionStrategy
     # Do not add property if the rdf_resource already responds to the property name
     # Do not add property if the rdf_resource already has a property with the same predicate.
     def self.apply(resource, property)
-      return if resource.respond_to?(property.name)
-      return if resource.properties.any? { |p| p[1].predicate == property.predicate }
-      resource.property property.name, property.to_h
+      return if resource.metadata_schema.respond_to?(property.name)
+      return if resource.metadata_schema.properties.any? { |p| p[1].predicate == property.predicate }
+      resource.metadata_schema.property property.name, property.to_h
```
grosscol: The properties are no longer defined directly on the resource?  What is metadata_schema?

lib/hydra/works/services/characterization_service.rb
awead: yes. Although, if content is nil, then you can't persist the tech data.

lib/hydra/works/models/characterization/already_there_strategy.rb
awead: resource here is the File object, which by LDP rules, is just the binary content. In order to define RDF properties, we need a RDF resource, which is what metadata_schema represents. In Fedora's terminology, this is fcr:metdata.

escowles: That makes sense -- though I think the name `metadata_schema` might be confusing.  Is that coming from ActiveFedora?  It might be nice to change it to `linked_description` or `binary_description`.

awead: Yes, that's AF. I can change that if we agree.

jcoyne: I'm of the opinion that `_schema` should stay, because it's describing the schema for description, not the description of any particular object.

jcoyne: I think we should change the name of this class to show that it only applies to pcdm:Files now.

escowles: @jcoyne I misread that as being the description of a binary, not the schema.  So `metadata_schema` here is one of the audio, video, document, etc. schemas?  Calling that `metadata_schema` sounds good to me.

lib/hydra/works/services/characterization_service.rb
```diff
@@ -63,9 +57,10 @@ def extract_metadata(content)
         end
       end
 
+      # Generate a random file name
       def temp_file_name
-        m = %r{/([^/]*)$} .match(object.uri)
-        "#{m[1]}-content.tmp"
+        m = (0...8).map { (65 + rand(26)).chr }.join
```
escowles: Can we just use the builtin Ruby tempfile function here?  Or is there a reason we need custom tempfile name generation?

awead: Originally, it was filename created using the uri. However, uri doesn't exist yet because the content hasn't been saved. Maybe it should just already exist? I can change the service around to raise an error if it doesn't... You can still pass in a file in order to avoid re-downloading the content from Fedora.

awead: Didn't know about Ruby's built-in tempfile, so 👍  to that if we want to use it instead.

lib/hydra/works/characterization/schema/base_schema.rb
```diff
@@ -10,8 +10,6 @@ class BaseSchema < ActiveTriples::Schema
     property :fits_version, predicate: RDF::Vocab::PREMIS.hasCreatingApplicationVersion
     property :exif_version, predicate: RDF::Vocab::EXIF.exifVersion
     property :original_checksum, predicate: RDF::Vocab::NFO.hashValue
-    property :mime_type, predicate: RDF::Vocab::EBUCore.hasMimeType, multiple: false do |index|
-      index.as :stored_searchable
-    end
+    property :mime_type, predicate: RDF::Vocab::EBUCore.hasMimeType
```
mjgiarlo: Does this no longer need to be indexed?

mjgiarlo:
Refs projecthydra/sufia#1894 

Fixes #290, #287

lib/hydra/works/characterization/schema/base_schema.rb
awead: @mjgiarlo the FileSet will be responsible for indexing it. Since this is Hydra::PCDM::File none of its terms are indexed.

lib/hydra/works/characterization.rb
```diff
@@ -14,12 +27,6 @@ module Characterization
       autoload :VideoSchema
     end
 
-    autoload :Base, 'hydra/works/models/concerns/file_set/characterization/base.rb'
-    autoload :Image, 'hydra/works/models/concerns/file_set/characterization/image.rb'
-    autoload :Document, 'hydra/works/models/concerns/file_set/characterization/document.rb'
-    autoload :Video, 'hydra/works/models/concerns/file_set/characterization/video.rb'
-    autoload :Audio, 'hydra/works/models/concerns/file_set/characterization/audio.rb'
-
     included do
```
jcoyne: I think you can remove this block now that nothing includes this as a module.

lib/hydra/works/characterization.rb
```diff
@@ -3,8 +3,21 @@ module Characterization
     extend ActiveSupport::Concern
```
jcoyne: Remove as this is no longer acts as a concern.

lib/hydra/works/services/characterization_service.rb
```diff
@@ -31,41 +33,26 @@ def characterize
...
       end
 
-      def temp_file_name
-        m = %r{/([^/]*)$} .match(object.uri)
-        "#{m[1]}-content.tmp"
+      # Determine the filename to send to Hydra::FileCharacterization. If no source is present,
+      # use the name of the file from the object; otherwise, use the supplied source.
+      def file_name
+        return "original_file" unless object.original_name
+        return object.original_name unless source
```
jcoyne: There are no tests to cover this line, or the next 2.

awead:
@mjgiarlo @jcoyne this is ready for a final review

PR #295, opened 2016-05-06
Extracting VirusCheckService from module mixin

jeremyf:
One of the major disadvantages of module mixins is that several methods
often get defined on the object. This greatly increases the chance of
method collisions if lots of mixins are used.

So, while preserving much of the outward interface of the
`Hydra::Works::VirusCheck`'s validation, I'm extracting a collaborating
class that can do the heavy lifting against the file.

This involved a bit of additional refactoring as I was encountering
challenges with ClamAV definitions. By relying also on dependency
injection of the system virus scanner, I can more easily test each
piece of the anti-virus process in better isolation.

### DIFF
--- a/lib/hydra/works/models/concerns/file_set/virus_check.rb
+++ b/lib/hydra/works/models/concerns/file_set/virus_check.rb
@@ -4,48 +4,12 @@ module Hydra::Works
 
     included do
       validate :detect_viruses
+      def detect_viruses
+        return true unless original_file && original_file.new_record? # We have a new file to check
+        return true unless VirusCheckerService.file_has_virus?(original_file)
+        errors.add(:base, "Failed to verify uploaded file is not a virus")
+        false
+      end
     end
-
-    # Default behavior is to raise a validation error and halt the save if a virus is found
-    def detect_viruses
-      return unless original_file && original_file.new_record?
-      path = original_file.is_a?(String) ? original_file : local_path_for_file(original_file)
-      unless defined?(ClamAV)
-        warning "Virus checking disabled, #{path} not checked"
-        return
-      end
-      scan_result = ClamAV.instance.scanfile(path)
-      handle_virus_scan_results(path, scan_result)
-    end
-
-    private
-
-      def handle_virus_scan_results(path, scan_result)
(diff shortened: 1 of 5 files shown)


lib/hydra/works/models/concerns/file_set/virus_check.rb
```diff
@@ -4,48 +4,12 @@ module VirusCheck
...
-
-    # Default behavior is to raise a validation error and halt the save if a virus is found
-    def detect_viruses
-      return unless original_file && original_file.new_record?
-      path = original_file.is_a?(String) ? original_file : local_path_for_file(original_file)
-      unless defined?(ClamAV)
-        warning "Virus checking disabled, #{path} not checked"
-        return
+      def detect_viruses
+        return true unless original_file && original_file.new_record? # We don't have a new file
```
jcoyne: reword this comment please. We *do* have a new file.

lib/hydra/works/services/virus_checker_service.rb
```diff
@@ -0,0 +1,73 @@
...
+  # with the underlying system virus scanner.
+  class VirusCheckerService
+    attr_accessor :original_file, :system_virus_scanner
+
+    # @api public
+    # @param original_file [String, #path]
+    # @return true if no virus is detected
+    # @return false if the virus scanner was not configured
+    # @return false if a virus was detected
+    def self.verify_file_is_not_a_virus?(original_file)
```
jcoyne: file_has_virus? Might be easier to understand. I don't like the "not" in the name

jcoyne:
This is a good refactor. Just a few issues and make the build green and it shall be merged

jeremyf:
@jcoyne I'm struggling with my Solr instance. I am now stuck in LDP::Gone exception hell. But I'll make the suggested changes.

lib/hydra/works/services/virus_checker_service.rb
jeremyf: I dislike negatives but was aiming for a similar return value. However, since it is an isolated class, I like your suggestion.

jcoyne:
Lovely 👏

PR #297, opened 2016-05-16
Use our own VirusScanner, fixes #296

### LINKED ISSUE #296: Virus checks should report "no viruses" if no scanner is present

awead:
Creates a settable virus scanner class that defaults to ClamAV, if it is present, or sets a null scanner that reports files have no viruses.

### DIFF
--- a/lib/hydra/works/services/virus_checker_service.rb
+++ b/lib/hydra/works/services/virus_checker_service.rb
@@ -6,14 +6,12 @@ module Hydra::Works
 
     # @api public
     # @param original_file [String, #path]
-    # @return true if a virus was detected
-    # @return true if anti-virus was not able to run
-    # @return false if the file was scanned and no viruses were found
+    # @return true or false result from system_virus_scanner
     def self.file_has_virus?(original_file)
       new(original_file).file_has_virus?
     end
 
-    def initialize(original_file, system_virus_scanner = default_system_virus_scanner)
+    def initialize(original_file, system_virus_scanner = Hydra::Works.default_system_virus_scanner)
       self.original_file = original_file
       self.system_virus_scanner = system_virus_scanner
     end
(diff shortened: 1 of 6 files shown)


lib/hydra/works/services/virus_checker_service.rb
```diff
@@ -27,30 +27,10 @@ def file_has_virus?
...
-        case scan_result
-        when 0 then return false
-        when 1
-          warning("A virus was found in #{path}: #{scan_result}")
-          true
-        else
-          warning "Virus checking disabled, #{path} not checked"
-          true
-        end
+        return false if scan_result == 0
```
mjgiarlo: The `0` here is a ClamAV thing. Should we instead expect the new `VirusScanner` class's `#call` method to return a boolean so that we don't need to bake scanner-specific logic into other code?

mjgiarlo: Also, now that this method will not be doing much "handling," but rather just returning the boolean it's passed (if you're cool with my :arrow_up: suggestion) and logging a message if a virus is found, I wonder if this method name should change to something simpler, like `#has_virus?`

lib/hydra/works/virus_scanner.rb
```diff
@@ -0,0 +1,32 @@
...
+# virus checker other than ClamAV, you'll need to configure it by creating your own class
+# that responds to ::call and setting in Hydra::Works. For example:
+#
+#   class MyScanner
+#     def self.call(file)
+#       Scanner.check_for_viruses(file)
+#     end
+#   end
+#
+#   Hydra::Works.default_system_virus_scanner = MyScanner
```
mjgiarlo: Do we want to recommend that folks set this in an initializer in Rails environments? Or meh? I could go either way on that.

lib/hydra/works/virus_scanner.rb
```diff
@@ -0,0 +1,32 @@
...
+module Hydra::Works
+  class VirusScanner
+    def self.call(file)
+      if defined?(ClamAV)
+        ClamAV.instance.method(:scanfile).call(file)
+      else
+        null_scanner(file)
+      end
+    end
```
mjgiarlo: Here's where I'd like to see some code that handles scanner-specific return values and turns them into booleans.

lib/hydra/works/virus_scanner.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    # Always return zero if there's nothing available to check for viruses. This means that
+    # we assume all files have no viruses because we can't conclusively say if they have or not.
+    def self.null_scanner(file)
+      if ActiveFedora::Base.logger
+        ActiveFedora::Base.logger.warn("Unable to check #{file} for viruses because no virus scanner is defined")
+      end
+      0
```
mjgiarlo: ClamAV smell. :nose:

lib/hydra/works/services/virus_checker_service.rb
awead: Yeah, I think virus scanner should have has_virus which returns true or false

lib/hydra/works/virus_scanner.rb
awead: Meh. Wherever. For Sufis and cc it'd be the init config I think

mjgiarlo: K. Retracting.

lib/hydra/works/virus_scanner.rb
```diff
@@ -0,0 +1,32 @@
...
+#       Scanner.check_for_viruses(file)
+#     end
+#   end
+#
+#   Hydra::Works.default_system_virus_scanner = MyScanner
+module Hydra::Works
+  class VirusScanner
+    def self.call(file)
+      if defined?(ClamAV)
+        ClamAV.instance.method(:scanfile).call(file)
```
jeremyf: You can switch this to `ClamAV.instance.scanfile(file)`; No need to detach the method.

lib/hydra/works/virus_scanner.rb
```diff
@@ -0,0 +1,56 @@
...
+      scan_result = ClamAV.instance.method(:scanfile).call(file)
+      return false if scan_result == 0
+      warning "A virus was found in #{file}: #{scan_result}"
+      true
+    end
+
+    # Always return zero if there's nothing available to check for viruses. This means that
+    # we assume all files have no viruses because we can't conclusively say if they have or not.
+    def null_scanner
+      warning "Unable to check #{file} for viruses because no virus scanner is defined"
```
mjgiarlo: This might annoy some folks, but I'm OK with leaving it in for now.

mjgiarlo:
:shipit:

awead:
I can fix the coveralls issue, if we want.

PR #299, opened 2016-05-24
Only use mime-type as defined on the file in ActiveFedora

awead:
This adjusts HydraWorks to align with recent changes in ActiveFedora regarding mime-type. It also addresses a small bug found with using this with Sufia. In passing a File object for upload, the file will have already been read by an ingest job and therefore needs to be rewound to the beginning in order to be read again for characterization.


awead:
@jcoyne ping.

PR #300, opened 2016-06-01
clarify virus detection method names

barmintor:
- detection method returns true if virus is detected
- custom ActiveModel validation is more idiomatic


escowles:
👍  -- definitely makes the virus checking less confusing.

PR #303, opened 2016-07-28
Updating hydra-pcdm dependency and adding test coverage for member_of_collections method on Works

escowles:


Gemfile
```diff
@@ -3,6 +3,9 @@ source 'https://rubygems.org'
 # Specify your gem's dependencies in hydra-works.gemspec
 gemspec
 
+gem 'active-fedora', github:'projecthydra/active_fedora', branch: 'master'
```
jcoyne: we probably shouldn't merge this PR until these are resolved to released versions.

escowles: @jcoyne definitely — I'm trying to get memberOf functionality in CurationConcerns today for discussion.  We'll want to have released versions of hydra-pcdm, hydra-works, and hydra-derivatives.

PR #305, opened 2016-08-22
Update ActiveFedora and require ActiveFedora::Datastreams needed for FitsDatastream

escowles:


jcoyne:
I prefer the solution in #306 to this.  Are you okay with that @escowles?

escowles:
Closing in favor of #306

PR #306, opened 2016-08-22
Use plain OM::XML::Document without the datastream

awead:
This supports ActiveFedora 11 and removes the dependency on OmDatastream.

### DIFF
--- a/hydra-works.gemspec
+++ b/hydra-works.gemspec
@@ -21,6 +21,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'hydra-pcdm', '>= 0.8'
   spec.add_dependency 'hydra-derivatives', '~> 3.0'
   spec.add_dependency 'hydra-file_characterization', '~> 0.3', '>= 0.3.3'
+  spec.add_dependency 'om', '~> 3.1'
 
   spec.add_development_dependency 'bundler', '~> 1.7'
   spec.add_development_dependency 'rake', '~> 10.0'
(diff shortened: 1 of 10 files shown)


awead:
ping @mjgiarlo @escowles

escowles:
This looks good to me — the build errors can probably be fixed with: https://github.com/projecthydra/hydra-works/blob/06d5cba043b1bfdb1e7570ec189cc446cc01e3fe/.travis.yml#L9-L10

hydra-works.gemspec
```diff
@@ -21,6 +21,8 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'hydra-pcdm', '>= 0.8'
   spec.add_dependency 'hydra-derivatives', '~> 3.0'
   spec.add_dependency 'hydra-file_characterization', '~> 0.3', '>= 0.3.3'
+  spec.add_dependency 'active-fedora', '~> 11.0.0.rc7'
```
mjgiarlo: Does this PR **require** AF 11, or would it work without this constraint? I'm concerned about forcing AF 11 on folks before we have tooling in place for them to deal with unordered values.

jcoyne: Do we need this? This will force everyone to use `rdf 2`.  Is there any reason it wouldn't work with version 10?

awead: I required AF 11 so I could reproduce the original missing constant AF::OmDatastream error, but a datastream-less implementation should work fine under AF 10.

awead:
@jcoyne I removed the AF constraint, although I haven't tested this with Sufia master, it should fix the OmDatastream error.

mjgiarlo:
:clap:

PR #308, opened 2016-09-02
Use services for determining mime type and filename

awead:
This will ultimately relate to fixing projecthydra/sufia#2553. We interrogate an incoming object about its original name and mime type, providing a values ffor each based on some logic. Testing this was difficult when it was buried in a private method. This surfaces the methods as services which can be more easily tested, as well as exposed to other consuming applications should they wish to make use of them.

### DIFF
--- a/lib/hydra/works/services/determine_mime_type.rb
+++ b/lib/hydra/works/services/determine_mime_type.rb
@@ -1,0 +1,39 @@
+module Hydra::Works
+  class DetermineMimeType
+    # Determines the mime type for a given file
+    # @param [IO, File, Rack::Multipart::UploadedFile, #read] file
+    # @param [String, NilClass] original_name of the file
+    # @return [String]
+    def self.call(file, original_name = nil)
+      new(file, original_name).determine_mime_type
+    end
+
+    attr_reader :file, :original_name
+
+    def initialize(file, original_name)
+      @file = file
+      @original_name = original_name
+    end
+
+    def determine_mime_type
+      return file.mime_type if mime_type?
+      return file.content_type if content_type?
+      mime_type_from_name_or_path || 'application/octet-stream'
+    end
+
+    def mime_type_from_name_or_path
+      return Hydra::PCDM::GetMimeTypeForFile.call(original_name) if original_name.present?
+      return Hydra::PCDM::GetMimeTypeForFile.call(file.path) if file.respond_to?(:path)
+    end
+
+    private
+
+      def mime_type?
+        file.respond_to?(:mime_type) && file.mime_type.present?
+      end
+
+      def content_type?
+        file.respond_to?(:content_type) && file.content_type.present?
+      end
+  end
+end
--- a/lib/hydra/works/services/determine_original_name.rb
+++ b/lib/hydra/works/services/determine_original_name.rb
@@ -1,0 +1,13 @@
+module Hydra::Works
+  class DetermineOriginalName
+    # Determines the original name for a given file
+    # @param [IO, File, Rack::Multipart::UploadedFile, #read] file
+    # @return [String]
+    def self.call(file)
+      return file.original_name if file.respond_to?(:original_name)
+      return file.original_filename if file.respond_to?(:original_filename)
+      return ::File.basename(file.path) if file.respond_to?(:path)
+      ''
+    end
+  end
+end
--- a/spec/hydra/works/services/add_file_to_file_set_spec.rb
+++ b/spec/hydra/works/services/add_file_to_file_set_spec.rb
@@ -67,10 +67,10 @@ describe Hydra::Works::AddFileToFileSet do
     end
   end
 
-  context 'file responds to :path but not to :mime_type nor :original_name' do
-    it 'defaults to Hydra::PCDM for mimetype and ::File for basename.' do
-      expect(Hydra::PCDM::GetMimeTypeForFile).to receive(:call).with(file.path)
-      expect(::File).to receive(:basename).with(file.path)
+  context 'when determining mime type and name' do
+    it 'uses services to assign the values' do
+      expect(Hydra::Works::DetermineOriginalName).to receive(:call).with(file)
+      expect(Hydra::Works::DetermineMimeType).to receive(:call).with(file, nil)
       described_class.call(file_set, file, type)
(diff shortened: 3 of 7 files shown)


lib/hydra/works/services/determine_mime_type.rb
```diff
@@ -0,0 +1,24 @@
+module Hydra::Works
+  class DetermineMimeType
+    # Determines the mime type for a given file
+    # @param [IO, File, Rack::Multipart::UploadedFile, #read] file
+    # @param [String, NilClass] original_name of the file
+    # @return [String]
+    class << self
```
mjgiarlo: I don't feel strongly about this, especially given we use different patterns in the hydra-works codebase, but I have a slight preference for the service object design that uses a class method to create an instance. Like so:

https://github.com/projecthydra/hydra-works/blob/master/lib/hydra/works/services/characterization_service.rb#L9-L21

That way you can make use of `attr_accessor` instead of passing around the same arguments (e.g., `file` and `original_name`). Like I said, slight preference.

lib/hydra/works/services/determine_mime_type.rb
```diff
@@ -0,0 +1,24 @@
+module Hydra::Works
+  class DetermineMimeType
+    # Determines the mime type for a given file
```
mjgiarlo: I believe this method documentation should sit above the `#call` method definition rather than above `class << self`.

lib/hydra/works/services/determine_original_name.rb
```diff
@@ -0,0 +1,15 @@
+module Hydra::Works
+  class DetermineOriginalName
+    # Determines the original name for a given file
+    # @param [IO, File, Rack::Multipart::UploadedFile, #read] file
+    # @return [String]
+    class << self
```
mjgiarlo: Same comments as above re: service object design and moving the method docs down a line.

lib/hydra/works/services/determine_mime_type.rb
```diff
@@ -0,0 +1,24 @@
...
+        return file.content_type if file.respond_to?(:content_type) && file.content_type.present?
+        mime_type_from_name_or_path(file, original_name) || default_mime_type
+      end
+
+      def default_mime_type
+        'application/octet-stream'
+      end
+
+      def mime_type_from_name_or_path(file, original_name)
+        return Hydra::PCDM::GetMimeTypeForFile.call(original_name) if original_name.present?
```
mjgiarlo: This looks like a new condition (not seeing this above). Is this a bug fix change in addition to the refactor?

spec/hydra/works/services/add_file_to_file_set_spec.rb
```diff
@@ -67,14 +67,6 @@
     end
   end
 
-  context 'file responds to :path but not to :mime_type nor :original_name' do
```
mjgiarlo: Is it worth changing this test to just make sure both of the new service objects are called (but otherwise defer testing said objects to the new specs below)?

lib/hydra/works/services/determine_mime_type.rb
awead: @mjgiarlo I was going by what's currently in the repo. The majority use `self.call` without an additional initializer, but I share your opinions. The primary motivation behind doing this was to avoid having to use Rspec's `allow_any_instance_of` because the original `determine_mime_type` method was nested deep in a private method, in an internal class. I  found it too difficult to mock that properly, and it highlighted the fact that `AddFileToFileSet` was too concerned with mime types and names.

True, it is basically a module, but in a service interface it's more accessible to another application. Of course this is a bit YAGNI, but like I said, the primary motivation for this was to get around mocking issues and adhering more to single-responsibility principles. I could have gone with the module design, but a service, I think, is an extra bonus in this case.

mjgiarlo: :ok_hand:

lib/hydra/works/services/determine_mime_type.rb
awead: @mjgiarlo it's going to relate to a forthcoming PR in projecthydra/curation_concerns@3cf65ca3352fad35d75e0f8505e623d4b45a0a6f that ultimately gets the original Sufia issue to pass. Do you see where I am here? Three layers down, and three gem releases just to fix one issue. 😉 

There are cases when we want to find the mime type using the provided original name, and _not_ the path. Examples are when my file has been uploaded as `sample.txt1234sdfasdf234`. The path doesn't help me there, so if the file is always represented as an object that responds to `original_name` or `original_filename`, I can supply the right name as opposed to having to rely on the path.

spec/hydra/works/services/add_file_to_file_set_spec.rb
awead: Seems reasonable. It integrates the service with the other two.

lib/hydra/works/services/determine_mime_type.rb
awead: FWIW, I would love to see some agreement on how we write our services and have some kind of automated enforcement that ensures we're consistent. Ruboservice anyone?

mjgiarlo: At the very minimum, it might be nice to arrive at two patterns:

* a Module-like pattern (a bunch of methods defined on `self`)
* a Class-like pattern (a single method defined on `self` named `#run` or `#call` that instantiates the class, uses attrs to encapsulate state, and assigns passed in params to ivars in `#initialize`) -- and it'd be great to standardize on either `#run` or `#call`

lib/hydra/works/services/determine_mime_type.rb
mjgiarlo: OK! I figured. :)

awead:
@mjgiarlo ok, fixed it up. Incidentally, I changed `DetermineMimeType` to use the call/initializer strategy. Upon doing that, rubocop started complaining that `Assignment Branch Condition size for determine_mime_type is too high` 😠  Necessitating private methods.

mjgiarlo:
@awead RUBOCOP. No good deed, eh? :clap:

awead:
thanks, @mjgiarlo !

PR #312, opened 2017-01-11
Changing to use stream instead of content so that the content is stre…

### LINKED ISSUE #311: Out of Memory issue for large files during virus check
This code cuases an out of memory error for large files since it reads the file entirely into memory and then uploads writes it to disk: https://github.com/projecthydra/hydra-works/blob/master/lib/hydra/works/services/virus_checker_service.rb#L33-L34

cam156:
…amed from fedora

Fixes #311 

@escowles does this work for you?   I know I did not pull in the CC code, but that got complicated really fast...

### DIFF
--- a/lib/hydra/works/services/virus_checker_service.rb
+++ b/lib/hydra/works/services/virus_checker_service.rb
@@ -30,8 +30,9 @@ module Hydra::Works
         return file.path if file.respond_to?(:path)
         Tempfile.open('') do |t|
           t.binmode
-          t.write(file.content.read)
-          file.content.rewind
+          file.stream.each do |chunk|
+            t.write(chunk)
+          end
           t.close
           t.path
         end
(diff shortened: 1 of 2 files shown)


lib/hydra/works/services/virus_checker_service.rb
```diff
@@ -30,8 +30,9 @@ def local_path_for_file(file)
         return file.path if file.respond_to?(:path)
         Tempfile.open('') do |t|
           t.binmode
-          t.write(file.content.read)
-          file.content.rewind
```
escowles: Does removing this change the state of the file at the end of virus checking?  Or do we not need to do that any more because streaming the file doesn't consume it the same way calling `content` did?

cam156: @escowles FileBody does not have a rewind, so I removed it.  https://github.com/projecthydra/active_fedora/blob/a641a8f0d65e7c01f395eeb58c5e87c638395e2e/lib/active_fedora/file/streaming.rb

I believe that each time your call the each method it reopens the connection to fedora:

https://github.com/projecthydra/active_fedora/blob/a641a8f0d65e7c01f395eeb58c5e87c638395e2e/lib/active_fedora/file/streaming.rb#L27

escowles: 👍

PR #314, opened 2017-01-18
Changing how the file is retrieved for virus checking

### LINKED ISSUE #313: VirusCheckerService tries to access stream of FileSet that has not been saved
The way the VirusCheckerService is being use in sufia/curation concerns, as a validation before save, causes the service to try and access a stream from an item that is not yet in the repository.

https://github.com/projecthydra/hydra-works/commit/796f8d12629b62442ac248e26cbe466e05e9828f#diff-8adc158121e9605cac64b6dc8574f678R30

The class of the content is a File in this case so we should be able to either use the existing code for items that do exists in the database `!item.new_record? ` or Check for the content responding to path.  I added `return file.content.path if file.content.respond_to?(:path)` in my local copy of the gem and that has fixed the problem.

I think the entire solution may be check the content for path, then check for new_record and return content and after that return the stream.

cam156:
fixes #313 
 
 If the content has a path just use that
 If the file is not saved to fedora use the content
 If the file is save use the datastream

This bug was introduced by me in an effort to not read large files into memory.  Unfortunately when the item was not saved the datastream could not be retrieved from fedora.

### DIFF
--- a/spec/hydra/works/services/virus_checker_service_spec.rb
+++ b/spec/hydra/works/services/virus_checker_service_spec.rb
@@ -5,9 +5,11 @@ describe Hydra::Works::VirusCheckerService do
   let(:file) { Hydra::PCDM::File.new }
   let(:virus_checker) { described_class.new(file, system_virus_scanner) }
   let(:datastream) { instance_double "ActiveFedora::File::Streaming::FileBody" }
+  let(:content) { instance_double "File" }
 
   before do
     allow(file).to receive(:stream).and_return(datastream)
+    allow(file).to receive(:content).and_return(content)
     allow(datastream).to receive(:each)
   end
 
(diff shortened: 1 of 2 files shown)


spec/hydra/works/services/virus_checker_service_spec.rb
```diff
@@ -38,17 +41,47 @@
...
+      before do
+        allow(file).to receive(:respond_to?).with(:path).and_return(false)
+      end
+      context 'that the content responds to path' do
+        let(:content) { instance_double "File", read: ["abc123", nil], rewind: true, path: "abc123" }
+
+        it 'will return true' do
+          expect(system_virus_scanner).to receive(:infected?).with(kind_of(String)).and_return(false)
+          expect(content).not_to receive(:read)
+          expect(virus_checker.file_has_virus?).to eq(false)
```
jeremyf: The "will return true" is in contradiction of the `expect(virus_checker.file_has_virus?).to eq(false)`

spec/hydra/works/services/virus_checker_service_spec.rb
```diff
@@ -38,17 +41,47 @@
...
+        end
+      end
+      context "that the content does not respond to path" do
+        let(:content) { instance_double "File", read: ["abc123", nil], rewind: true }
+        it 'will return true' do
+          expect(system_virus_scanner).to receive(:infected?).with(kind_of(String)).and_return(false)
+          expect(file).not_to receive(:stream)
+          expect(virus_checker.file_has_virus?).to eq(false)
```
jeremyf: The "will return true" is in contradiction of the `expect(virus_checker.file_has_virus?).to eq(false)`

spec/hydra/works/services/virus_checker_service_spec.rb
cam156: @jeremyf good catch!  I'll change it at line 47 too since that is where I copy pasted it from.

cam156:
@jeremyf I updated the test names in the three locations that should be connected.

spec/hydra/works/services/virus_checker_service_spec.rb
jeremyf: @cam156 can you review the `it` and expectation

cam156:
@jeremyf Looks like the original test was 100% backwards on the it will return true vs it will return false.   Looked at them all and switched them to match the expects.

PR #319, opened 2017-02-08
Allows adding URL redirects as files

afred:
This creates a service class called AddExternalFileToFileSet which acts very
similar to AddFileToFileSet, but instead of expecting binary data, it expects a
URL. This URL will go into the external-content header of the Fedora resource,
and Fedora will automatically redirect to that URL.

The motivation is to provide the basis for more advanced external storage
features higher up in the Hydra stack.

### DIFF
--- a/lib/hydra/works.rb
+++ b/lib/hydra/works.rb
@@ -48,6 +48,7 @@ module Hydra
     autoload_under 'services' do
       autoload :VirusCheckerService
       autoload :AddFileToFileSet
+      autoload :AddExternalFileToFileSet
       autoload :UploadFileToFileSet
       autoload :PersistDerivative
       autoload :CharacterizationService
--- a/lib/hydra/works/services/add_external_file_to_file_set.rb
+++ b/lib/hydra/works/services/add_external_file_to_file_set.rb
@@ -1,22 +1,26 @@
 module Hydra::Works
-  class AddFileToFileSet
+  class AddExternalFileToFileSet
     # Adds a file to the file_set
     # @param [Hydra::PCDM::FileSet] file_set the file will be added to
-    # @param [IO,File,Rack::Multipart::UploadedFile, #read] object that will be the contents. If file responds to :mime_type, :content_type, :original_name, or :original_filename, those will be called to provide metadata.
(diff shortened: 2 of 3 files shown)


lib/hydra/works/services/add_file_to_file_set.rb
```diff
@@ -76,6 +76,7 @@ def find_or_create_file_for_rdf_uri(type, update_existing)
             current_file = file_set.files.last
             Hydra::PCDM::AddTypeToFile.call(current_file, type_to_uri(type))
           end
+          current_file
```
mjgiarlo: Can you say more about this change?

afred: Yes, tl;dr - it ensures the return value of `current_file` (instead of `nil` as we were getting at first).

Consider this simplified logic:
```ruby
def foo(x)
  y = "x is true" if x
  unless x
    "x is not true"
  end
end

foo(false)
# => "x is false"
foo(true)
# => nil
```

So... we started by copying `#find_or_create_file_for_rdf_uri` from `AddFileToFileSet::Updater` to our new `AddExternalFileToFileSet::Updater`, without the explicit return value.

Then, in our testing context, we were adding new external file URLs to the same `FileSet` objects with the same RDF type.

The result, is that the first assignment worked, but the method was returning the results of the `unless` block, which is `nil` in that case. That value then get's assigned to the `@current_file` attribute of the `Updater` class, which broke stuff.

The takeaway is that `#find_or_create_file_for_rdf_uri` needs to either return a `File` object, or raise an error, but we found a case (in our tests) where it was returning `nil`.

Explicitly specifying the return value helps prevent bugs that we haven't discovered yet, because there are currently no unit tests around `AddFileToFileSet::Updater`. I would be in favor of adding them, but I was hoping that could be a separate ticket, since it's only tangentially related to what we're actually trying to add here.

afred: ^ helps if my simplified logic makes sense. Edited.

mjgiarlo: OK, I'm cool with that, and the test suite seems happy. Thanks!

PR #320, opened 2017-02-28
Removing content read from file input to save memory

cam156:
This change reverts a small portion of a larger change. https://github.com/projecthydra/hydra-works/commit/6e0dd6cc27ea64af816049b936515a4ef62ddc9b

When you run read it reads the entire content into memory.  This is not necessary since the File can be passed directly to Hydra::FileCharacterization


awead:
@cam156 LGTM. Anyone else?

PR #322, opened 2017-03-16
Removing an exclusion for a non-existent file

jeremyf:
```ruby
require 'psych'
['.rubocop.yml', '.rubocop_todo.yml'].each do |filename|
begin
  yaml = Psych.load_file(filename)

  def read(node)
    if node.is_a?(Hash)
      node.each_pair do |key, value|
        read(value)
      end
    elsif node.is_a?(Array)
      node.each do |value|
        read(value)
      end
    elsif node.is_a?(String)
      return unless node =~ /.rb\Z/
      return if node =~ /\*/
      return if File.exist?(node)
      puts node
    end
  end
  read(yaml)
rescue
end
end
```

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -7,7 +7,6 @@ AllCops:
     - '**/Rakefile'
   Exclude:
     - 'vendor/**/*'
-    - 'spec/internal/bin/*'
 
 Metrics/BlockLength:
   Exclude:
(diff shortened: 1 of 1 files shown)


.rubocop.yml
```diff
@@ -8,7 +8,6 @@ AllCops:
   Exclude:
     - 'vendor/**/*'
     - 'spec/internal/bin/*'
```
jcoyne: I'll bet that `spec/internal` is entirely gone since we've upgraded rubocop.  Want to remove that line too?

jeremyf:
@jcoyne Done

PR #330, opened 2017-06-23
Only store largest height, width for characterization, fixes #327

### LINKED ISSUE #327: Largest width / height should be used
Currently if a multi-layer tiff is characterized, fits returns multiple imageHeight and imageWidth entries with status="CONFLICT", e.g.
```
      <imageWidth toolname="Jhove" toolversion="1.5" status="CONFLICT">2226</imageWidth>
      <imageWidth toolname="Tika" toolversion="1.3" status="CONFLICT">160</imageWidth>
      <imageHeight toolname="Jhove" toolversion="1.5" status="CONFLICT">1650</imageHeight>
      <imageHeight toolname="Tika" toolversion="1.3" status="CONFLICT">119</imageHeight>
```

Current behavior is to just save both on the model.
```
 :width=>["2226", "160"],
 :height=>["1650", "119"],
```

Downstream apps don't appear to be accounting for this, which is their problem, sure.
e.g. Sufia's file page reports my object as having dimensions:
```
Height: 1650
Width: 160
```

But does it make any sense to have multiple heights / multiple widths, especially when they aren't paired appropriately? My preference would be to choose the largest of each before storing on the model https://github.com/samvera/hydra-works/blob/master/lib/hydra/works/services/characterization_service.rb#L31. Once it's on the model it's accessible in our apps as linked data and it's confusing.

hackmastera:
This can be reviewed; please don't merge until i've run it by the tech list as promised on this week's call.
Update: this is ready for review / merge.

### DIFF
--- a/lib/hydra/works/services/characterization_service.rb
+++ b/lib/hydra/works/services/characterization_service.rb
@@ -101,7 +101,10 @@ module Hydra::Works
       end
 
       def append_property_value(property, value)
+        # We don't want multiple mime_types; this overwrites each time to accept last value
         value = object.send(property) + [value] unless property == :mime_type
+        # We don't want multiple heights / widths, pick the max
+        value = value.map(&:to_i).max.to_s if property == :height || property == :width
         object.send("#{property}=", value)
       end
   end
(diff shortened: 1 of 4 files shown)


cam156:
@hackmastera my only concern with this implementation is that you may or may not be matching up the height and width.  Is it possible for the arrays to be height: [5,25] & width: [7,2], so you would end up with height:25 width:7 ? I'm guessing that may never happen, but I thought I would point it out...

hackmastera:
@cam156 I'm not sure that such a file exists. however, if it did, I would probably argue that its cumulative dimensions are indeed 7 x 25.

hackmastera:
no idea what this failure means and I'm not getting it locally. @samvera/hyrax-code-reviewers?

mjgiarlo:
@hackmastera hmm, can you confirm that you've got AF 11.4.0 locally?

hackmastera:
@mjgiarlo good catch! I was running on 11.3.1; didn't update my bundle in the weeks between submitting this PR and coming back around to it. i'm getting the failures now after updating.

It looks like changing these specs from `new` to `create` fixes all but one failure. I don't have time to look at it more right now.

hackmastera:
Actually that one failure is one I had before since I don't have the dependency required for the docx thumnbnail generation. I'm going to push now and see what travis says about changing these from `new` to `create`.

hackmastera:
this is passing now. quick somebody review it before dependencies break it again! 😆  @samvera/hyrax-code-reviewers

mjgiarlo:
:clap:

PR #333, opened 2017-08-30
Test reversed nested collection relationships

cjcolvar:
This PR mirrors the one made by @escowles (#303) when reversing the collection->object relationship.  All of the work happens in hydra-pcdm so only tests need to be added here.  This PR points to the branch of hydra-pcdm which has the changes, but it should use a released version when available.


cjcolvar:
Depends on https://github.com/samvera/hydra-pcdm/pull/244.

PR #335, opened 2018-02-10
Fix #334 - Add a cleanup phase to CharacterizationService

### LINKED ISSUE #334: Characterization with FITS can report multiple mime types, which can break the File
When ingesting certain types of files, the characterization by FITS can report multiple mime types as a comma-separated attribute. The CharacterizationService document does not detect or clean this, nor does the OM document.

Once the File is updated and persisted to Fedora, the invalid mime type results in 500 errors thrown for any GET or HEAD requests for the file (with Jersey throwing a parsing exception on the comma).

It's reasonable to say that both FITS (or Droid) and Fedora are behaving badly in this case, but it is also reasonable to be able to clean up or transform values after extraction. OM does not have a mechanism for doing transformation of a term beyond the hoisting done with proxied terms, so a new method on the FitsDocument is the best candidate.

A snippet showing the problematic mimetype attribute:
```xml
<fits xmlns="http://hul.harvard.edu/ois/xml/ns/fits/fits_output" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://hul.harvard.edu/ois/xml/ns/fits/fits_output http://hul.harvard.edu/ois/xml/xsd/fits/fits_output.xsd" version="1.2.0" timestamp="2/7/18 10:03 PM">
  <identification status="SINGLE_RESULT">
    <identity format="netCDF-3 Classic" mimetype="application/netcdf, application/x-netcdf" toolname="FITS" toolversion="1.2.0">
      <tool toolname="Droid" toolversion="6.3" />
      <externalIdentifier toolname="Droid" toolversion="6.3" type="puid">fmt/282</externalIdentifier>
    </identity>
  </identification>
```

botimer:
This is a little suboptimal because OM exposes all of the terms on the
top level of the instance, so the method is named __cleanup__. The
parsers also do not have a base class to ensure that a given cleanup
method is available, so we use a respond_to? check. The last ugly bit is
that calling `file_mime_type = ...` did not update the value, so the
`send("filetype=", ...)` was the only way I could get this to pass.

There is a new test with FITS output from running against a sample
netCDF file. There may be other types that report multiple mimetypes in
this way, so the cleanup implementation is general (split on comma, pick
the first). It's possible that using an array would be better, but there
are other expectations in the service that enforce a single value.

### DIFF
--- a/lib/hydra/works/characterization/fits_document.rb
+++ b/lib/hydra/works/characterization/fits_document.rb
@@ -146,6 +146,20 @@ module Hydra::Works::Characterization
       t.frame_rate(proxy: [:metadata, :video, :frame_rate])
     end
 
+    # Cleanup phase; ugly name to avoid collisions.
+    # The send construct here is required to fix up values because the setters
+    # are not defined, but rather applied with method_missing.
+    def __cleanup__
+      # Sometimes, FITS reports the mimetype attribute as a comma-separated string.
+      # All terms are arrays and, in this case, there is only one element, so scan the first.
+      if file_mime_type.present? && file_mime_type.first.include?(',')
+        send("file_mime_type=", [file_mime_type.first.split(',').first])
+      end
+
+      # Add any other scrubbers here; don't return any particular value
+      nil
+    end
+
     def self.xml_template
       builder = Nokogiri::XML::Builder.new do |xml|
         xml.fits(xmlns: 'http://hul.harvard.edu/ois/xml/ns/fits/fits_output',
--- a/lib/hydra/works/services/characterization_service.rb
+++ b/lib/hydra/works/services/characterization_service.rb
@@ -62,6 +62,7 @@ module Hydra::Works
       def parse_metadata(metadata)
         omdoc = parser_class.new
         omdoc.ng_xml = Nokogiri::XML(metadata) if metadata.present?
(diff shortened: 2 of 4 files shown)


lib/hydra/works/characterization/fits_document.rb
```diff
@@ -146,6 +146,14 @@ class FitsDocument
       t.frame_rate(proxy: [:metadata, :video, :frame_rate])
     end
 
+    # Cleanup phase; ugly name to avoid collisions
+    def __cleanup__
```
mjgiarlo: What would you think of a couple of small changes to this method? Namely, something like:

```ruby
def __cleanup__
  # Check for conditions in a guard clause, in one fell swoop rather than two checks
  #   and don't bother modifying `file_mime_type` if it isn't in the problematic state (with a comma)
  return if file_mime_type.blank? || !file_mime_type.first.include?(',')
  # Not sure if `#send` is needed, but I'm A-OK with this method using it instead of the setter if so
  file_mime_type = [file_mime_type.first.split(','),first]
end
```

mjgiarlo:
Thanks for working on this, @botimer!

botimer:
Thanks for the review, @mjgiarlo. I've updated the PR with a bit of commentary and a style change. Definitely better now. See what you think!

mjgiarlo:
Thanks, @botimer!

Issue #337, opened 2018-02-21
Multiple, conflicting values for Exif Version in characterization metadata


kefo:
We have more than 200,000 pcdm:Files in our collection with multiple and unequal `exif:exifVersion` values.

There are, of course, multiple values in the `fits` output:

```xml
      <exifVersion toolname="Exiftool" toolversion="10.00" status="CONFLICT">0230</exifVersion>
      <exifVersion toolname="NLNZ Metadata Extractor" toolversion="3.6GA" status="CONFLICT">0320</exifVersion>
```
(I actually see 3 exifVersion values in our data sometimes and, for the life of me, I cannot figure out whence the third comes, but, again, I digress.)

Given the CONLICTing values above, I'm not sure what to do about this other than raise it as an issue.

In so far as this results in duplicate, conflicting values in the characterization metadata, it is related to https://github.com/samvera/hydra-works/issues/336.

kefo:
The most recent version of EXIF is 2.31, which would mean that NLNZ's value in the above snippet is incorrect.  It's probably not taking into account endianness, but, again, I digress.

Assuming that to be the base, is the best solution to rely on ExifTool's value? 

Something like:

~~~ruby
 t.exif_version(path: 'exifVersion', attributes: { toolname: "Exiftool" })
~~~

See:

https://github.com/samvera/hydra-works/blob/v0.17.0/lib/hydra/works/characterization/fits_document.rb#L62

jrgriffiniii:
Resolved with https://github.com/samvera/hydra-works/pull/338

Issue #336, opened 2018-02-21
Multiple, conflicting values for dateCreated in characterization metadata


kefo:
We have more than 200,000 pcdm:Files (most are images probably but I can't confirm this) in our collection with multiple and unequal `ebucore:dateCreated` values.

It's easy to see how this is happening when comparing the `fits` output:

```xml
  <fileinfo>
    <size toolname="Jhove" toolversion="1.16">862707</size>
    <creatingApplicationName toolname="Exiftool" toolversion="10.00" status="CONFLICT">Canon EOS 5D Mark III</creatingApplicationName>
    <creatingApplicationName toolname="NLNZ Metadata Extractor" toolversion="3.6GA" status="CONFLICT">Adobe Photoshop Lightroom 6.7 (Macintosh)</creatingApplicationName>
    <lastmodified toolname="Exiftool" toolversion="10.00" status="CONFLICT">2017:05:30 15:07:03</lastmodified>
    <lastmodified toolname="Tika" toolversion="1.10" status="CONFLICT">2017-05-30T15:07:03</lastmodified>
    <created toolname="Exiftool" toolversion="10.00" status="CONFLICT">2017:05:26 09:43:42</created>
    <created toolname="NLNZ Metadata Extractor" toolversion="3.6GA" status="CONFLICT">2017:05:30 15:07:03</created>
    <filepath toolname="OIS File Information" toolversion="0.2" status="SINGLE_RESULT">/Users/kford1/Work/fits/samples/image/IM023744_016.jpg</filepath>
    <filename toolname="OIS File Information" toolversion="0.2" status="SINGLE_RESULT">IM023744_016.jpg</filename>
    <md5checksum toolname="OIS File Information" toolversion="0.2" status="SINGLE_RESULT">2e887492065e262fc866a2028c160f4b</md5checksum>
    <fslastmodified toolname="OIS File Information" toolversion="0.2" status="SINGLE_RESULT">1519246100000</fslastmodified>
  </fileinfo>
```

to the code that parses the fits document, line 26:

https://github.com/samvera/hydra-works/blob/d41be82540eca9d3a76301df3dcea5b74c9d5937/lib/hydra/works/characterization/fits_document.rb#L23-L26

[...]

PR #338, opened 2018-03-30
PR to address Issues 336 337

kefo:
This addresses issues #336 and #337 .


kefo:
@escowles I'm gently tagging you only because you were kind enough to comment on the issue in the first place.  Let me know if you see a problem with this...

escowles:
@kefo This looks good to me — my only question is whether we always have dates from ExifTool.  Assuming we do always have them, then using the dates from the better tool sounds like the way to go.

escowles:
@kefo I would have expected it to fall back on the file creation/modification timestamps from the filesystem.  But that test makes me pretty confident that if there are dates, then ExifTool will have dates, and this won't be hiding info from other tools.

kefo:
@escowles That's exactly what I expected to happen too, which was my motive behind making sure the file contained valid EXIF data, checking what FITS reported, and then stripping the EXIF to see the new result in FITS.

PR #340, opened 2018-04-24
Av cherrypick

yingfeng-iu:
Avalon-bundle Issue#2915: Extending Technical Metadata in Generic Work Type

### DIFF
--- a/spec/hydra/works/services/characterization_service_spec.rb
+++ b/spec/hydra/works/services/characterization_service_spec.rb
@@ -158,7 +158,7 @@ describe Hydra::Works::CharacterizationService do
     end
 
     context 'using video metadata' do
-      let(:fits_filename) { 'fits_1.2.0_avi.xml' }
+      let(:fits_filename) { 'fits_0.8.5_avi.xml' }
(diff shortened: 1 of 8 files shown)


spec/hydra/works/services/characterization_service_spec.rb
```diff
@@ -81,20 +81,25 @@
...
-    let!(:current_schemas) { ActiveFedora::WithMetadata::DefaultMetadataClassFactory.file_metadata_schemas }
-
     let(:characterization) { class_double("Hydra::FileCharacterization").as_stubbed_const }
     let(:fits_filename)    { 'fits_0.8.5_pdf.xml' }
     let(:fits_response)    { IO.read(File.join(fixture_path, fits_filename)) }
     let(:file_content)     { 'dummy content' }
     let(:file)             { Hydra::PCDM::File.new { |f| f.content = file_content } }
 
-    before do
+    before(:all) do
```
cjcolvar: It looks like there is a rubocop issue with this construction so instead how about combining the before and after blocks into an [around block](https://relishapp.com/rspec/rspec-core/v/3-7/docs/hooks/around-hooks) ?
```
around do |example|
  ...
  example.run
  ...
end
```

PR #346, opened 2018-07-16
Replace OM with nokogiri-happymapper

jcoyne:


no-reply:
Yes, yes, yes!

I'd love to get this in prior to https://github.com/samvera/hydra-works/issues/343.

cjcolvar:
I'd like there to be more investigation and discussion before this gets merged so we can be very deliberate about our decision.  I feel like there is still a fair bit of XML out in the community which was/is being parsed and generated via OM (within Avalon's ecosystem there definitely is).  We still get occasional questions from people trying to work through the XML model lesson (https://github.com/samvera/hydra/wiki/Lesson---Build-a-Codex-model-with-XML).  It would be great if we could adopt a single approach for read and writing across samvera repos then update our documentation accordingly so we can provide some guidance.

This PR seems like one approach.  I'd like to see how nokogiri-happymapper work for generating xml especially the tricky scenarios in OM like when you need to generate multiple nodes for a value via templates.  

For what its worth [nom-xml](https://github.com/cbeer/nom-xml) is another option.

jcoyne:
I don't care. The main goal was to get rid of OM. if #354 does that 👍

mark-dce:
@jcoyne thanks!  I'm gonna close this for now then.

PR #358, opened 2019-01-31
Add parent and child methods that aggregate both directions of relationships

elrayle:


lib/hydra/works/models/concerns/collection_behavior.rb
```diff
@@ -44,6 +68,26 @@ def work_ids
...
     end
 
+    def member_collections
+      return [] if id.nil?
+      member_objects = ActiveFedora::Base.where('member_of_collection_ids_ssim' => id)
+      member_objects.select(&:collection?).to_a
+    end
+
+    def member_collection_ids
+      member_collections.map(&:id)
```
cam156: @elrayle Lets change this over to a solr query

lib/hydra/works/models/concerns/collection_behavior.rb
```diff
@@ -44,6 +68,26 @@ def work_ids
...
+    end
+
+    def member_works
+      return [] if id.nil?
+      member_objects = ActiveFedora::Base.where('member_of_collection_ids_ssim' => id)
+      member_objects.select(&:work?).to_a
+    end
+
+    def member_work_ids
+      member_works.map(&:id)
```
cam156: Here too.

PR #372, opened 2019-10-21
Deprecate clamav gem in favor of Clamby

cjcolvar:
Fixes #357 

Also had to pin sprockets to < 4 and solr to 7.x.


README.md
```diff
@@ -43,7 +43,14 @@ Behaviors included in the model include:
...
 
-Check out the [Hydra::Derivatives README](https://github.com/samvera/hydra-derivatives#dependencies) for additional dependencies.
+## Dependencies
+
+Check out the [Hydra::Derivatives README](https://github.com/samvera/hydra-derivatives#dependencies) for dependencies.
+
+## Additional dependencies required for specs
+
+* ClamAV
+  * [Mac installation](https://gist.github.com/gagarine/9168c1b7e4b5f55cb3254582e30d808e)
```
escowles: Can we just add the directions to the README instead of linking to a gist?

```
$ brew install clamav
$ cp /usr/local/etc/clamav/freshclam.conf.sample /usr/local/etc/clamav/freshclam.conf
$ freshclam
```

cjcolvar: Updated with that change.

PR #375, opened 2020-01-21
Adding Ruby 2.7.z and Rails 6.y.z releases to the CircleCI build configuration

### LINKED ISSUE #373: Add support for Ruby 2.7.z releases
Ruby 2.7.0 was released on 12/25/2019, and in accordance with the charter of the current phase of the Component Maintenance Working Group, the CircleCI configuration for this should be updated.

jrgriffiniii:
Resolves #373 and #374

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,10 +46,6 @@ workflows:
           ruby_version: 2.5.7
           rails_version: 6.0.2
       - test:
-          name: ruby2-4_rails6-0
-          ruby_version: 2.4.9
-          rails_version: 6.0.2
-      - test:
           name: ruby2-7_rails5-2
           ruby_version: 2.7.0
           rails_version: 5.2.4
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,45 +35,45 @@ workflows:
     jobs:
       - test:
           name: ruby2-7_rails6-0
-          ruby_version: 2.7.0
-          rails_version: 6.0.2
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -33,27 +33,51 @@ workflows:
...
+          ruby_version: 2.6.5
+          rails_version: 6.0.2
+      - test:
+          name: ruby2-5_rails6-0
+          ruby_version: 2.5.7
+          rails_version: 6.0.2
+      - test:
+          name: ruby2-4_rails6-0
+          ruby_version: 2.4.9
+          rails_version: 6.0.2
```
cjcolvar: Ruby 2.4 isn't supported by rails 6 so this job can be removed.

jrgriffiniii:
I suspect that the the build errors were due to a dependency conflict with bixby and RuboCop.

cjcolvar:
@jrgriffiniii The dependency issue is that `activesupport` is restricted to `< 6.0` in the gemspec.  Upping that to `< 6.1` or `< 7.0` might work but it might just reveal the other samvera dependencies have the same issue.
Looks like you're going to work on this next week so don't feel rushed to do anything before then.

jrgriffiniii:
https://github.com/samvera/hydra-works/pull/375/commits/12b3bf35c17126f244c17f3e3db89cf08dc82dea#diff-1be913fda154ff735569b8603e55f289R23 seems to be the only dependency blocking this now.

PR #383, opened 2021-12-03
CharacterizationService: teach #file_name about Valkyrie objects

dunn:


dunn:
not sure what to do about these failures https://app.circleci.com/pipelines/github/samvera/hydra-works/44/workflows/ebc144f3-9488-4329-a1d2-0942228063de/jobs/271

lib/hydra/works/services/characterization_service.rb
```diff
@@ -52,10 +52,15 @@ def extract_metadata(content)
...
-        else
-          object.original_name.nil? ? "original_file" : object.original_name
+        case object
+        when Hydra::PCDM::File
+          if source
+            source.is_a?(File) ? File.basename(source.path) : File.basename(source)
+          else
+            object.original_name.nil? ? "original_file" : object.original_name
+          end
+        when Hyrax::FileMetadata
```
no-reply, line 62: in the short term, would it be helpful to (also?) add an alias `#original_name` to `Hyrax::FileMetadata`?

dunn: we could; i guess in that case maybe we'd want to flip the branch order and check for `object#original_name` first, then fall back on source

the problem is that the object we provide as `source` errors on both branches of `source.is_a?(File) ? File.basename(source.path) : File.basename(source)` so that codepath is only appropriate when it's `Hydra::PCDM::File`

(i guess a further question is why it's preferred over the `#original_name` in the first place?)

PR #396, opened 2024-10-02
Testing for Ruby 3.3 and Rails 7.2

randalldfloyd:
Required re-doing the Solr and Fedora setup because something is wrong with the Solr core creation in the samvera/install_solr_core orb. This also required a local Solr configuration directory to pass into the core creation API; the configuration was taken directly from the one being used in [active_fedora](https://github.com/samvera/active_fedora/tree/main/lib/generators/active_fedora/config/solr/templates/solr/conf).


cjcolvar (approved):
Looks good!

PR #399, opened 2025-11-07
Adding support for Rails 8

randalldfloyd:
Will require a release of hydra-derivatives per https://github.com/samvera/hydra-derivatives/pull/267

### DIFF
--- a/hydra-works.gemspec
+++ b/hydra-works.gemspec
@@ -21,7 +21,6 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
 
   spec.add_dependency 'activesupport', '>= 5.2', '< 9.0'
-  spec.add_dependency 'concurrent-ruby', '1.3.4'
   spec.add_dependency 'hydra-derivatives', '>= 3.6'
   spec.add_dependency 'hydra-file_characterization', '~> 1.0'
   spec.add_dependency 'hydra-pcdm', '>= 0.9'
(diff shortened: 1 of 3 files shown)


hydra-works.gemspec
```diff
@@ -20,7 +20,8 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
-  spec.add_dependency 'activesupport', '>= 5.2', '< 8.0'
+  spec.add_dependency 'activesupport', '>= 5.2', '< 9.0'
+  spec.add_dependency 'concurrent-ruby', '1.3.4'
```
cjcolvar: Does this have to be pinned to a specific version?

randalldfloyd: Are you referring to `concurrent-ruby`?  This resolves the issue we've seen elsewhere, where doing a require on `hydra/works` causes:
```
NameError:
  uninitialized constant ActiveSupport::LoggerThreadSafeLevel::Logger
```
I tried removing that pin one more time to verify that it [still causes the suite to fail](https://app.circleci.com/pipelines/github/samvera/hydra-works/1441/workflows/3b03cfb5-96ce-4395-8870-112b83297312/jobs/14640).

cjcolvar: I think to deal with this problem in other places I decided to drop testing on rails < 7.2 since they are EOL.  What do you think about doing that here instead of pinning to an older `concurrent-ruby`?

randalldfloyd: @cjcolvar Removed EOL Rails from test matrix and tests are all green without pinning `concurrent-ruby`.

cjcolvar (approved):
Looks good!

